PR #23, opened 2020-12-03
Frames

danmatichuk:


travitch (commented):
Here is a first round.  I still need to look at MemTrace and Verification in a bit more detail

src/Pate/Verification.hs
```diff
@@ -178,18 +189,23 @@ verifyPairs logAction elf elf' blockMap dcfg pPairs = do
       liftIO . putStr $ ppEquivalenceStatistics stats
       return $ equivSuccess stats
 
+---------------------------------------------
+-- Top-level loop
+
 -- | Verify equivalence of the given pairs, as well as any
 -- resulting pairs that emerge
 runVerificationLoop ::
```
travitch, line 197: Can this get a high-level description of the approach it is taking? It looks like some kind of worklist - is there some required structure to the order in which goals are discharged?

src/Pate/Verification.hs
```diff
@@ -110,37 +119,44 @@ verifyPairs logAction elf elf' blockMap dcfg pPairs = do
...
+  (oMain, oPfm)  <- runDiscovery elf
+  (pMain, pPfm) <- runDiscovery elf'
 
-  liftIO $ LJ.writeLog logAction (PE.LoadedBinaries (elf, pfmO) (elf', pfmP))
+  liftIO $ LJ.writeLog logAction (PE.LoadedBinaries (elf, oPfm) (elf', pPfm))
 
   Some gen' <- liftIO N.newIONonceGenerator
-  let pfeats = W4PF.useBitvectors .|. W4PF.useSymbolicArrays
+  let pfeats = W4PF.useBitvectors .|. W4PF.useSymbolicArrays .|. W4PF.useIntegerArithmetic .|. W4PF.useStructs
   CBO.withYicesOnlineBackend W4B.FloatRealRepr gen' CBO.NoUnsatFeatures pfeats $ \sym -> do
```
travitch, line 129: I don't recall - do we take advantage of the online backend? Would it simplify anything to use the offline backend? We don't have to do it now, but knowing would be useful (e.g., we might be able to parallelize)

src/Pate/Verification.hs
```diff
@@ -250,377 +288,67 @@ printResult (Left err) = liftIO $ putStr . ppEquivalenceError $ err
...
+
+  -- first try proving equivalence by assuming that exact equality
+  -- is the only condition we are propagating backwards, so we
+  -- don't do any work to try to intelligently narrow this down
+
+  -- TODO: it's unclear how fine-grained to make this, or if it's even
+  -- a good idea, but it makes the tests finish in about 1/3 the time
+  result <- manifestError $
+    local (\env -> env { envPrecondProp = PropagateExactEquality }) $
+    provePostcondition pPair postcondSpec
```
travitch, line 315: It would be interesting to collect statistics on how many blocks are exactly equal

src/Pate/Verification.hs
```diff
@@ -929,110 +533,546 @@ regStateToAsn regs = do
...
+        
+  case result of
+    Just result' -> throwHere $ InequivalentError result'
+    _ -> return (eqInputsPred, eqInputs)
+
+--------------------------------------------------------
+-- Propagating preconditions
+
+-- | Guess a sufficient memory domain that will cause the
+-- given postcondition to be satisfied on the given equivalence relations
```
travitch, line 706: What sorts of things go into a guess? Also, the guess is just used as an optimization and is not required to be correct?

src/Pate/Memory/MemTrace.hs
```diff
@@ -947,55 +1200,186 @@ memOpFootprint (MemOp ptr dir cond w _ _) = MemFootprint ptr w dir cond
...
+    Just True -> r
+    Just False -> return $ falsePred sym
+    Nothing -> do
+      br <- r
+      andPred sym bl br
+
+
+-- | Is this read shadowed by a previous write?
+-- FIXME: there are tradeoffs here with respect to how much we want to
+-- hammer the solver asking this question precisely.
```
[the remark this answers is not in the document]
danmatichuk: this was actually stale code from an old implementation that I've now removed

src/Pate/Verification.hs
danmatichuk: I added a comment about this. It's definitely possible but not completely trivial to swap to an offline backend.

src/Pate/Verification.hs
danmatichuk: I agree, but currently this is only at the granularity of functions that were explicitly given (i.e. the 'main' function). So this test is just checking if that top-level entry point can be proven equivalent without any intelligent precondition propagation, and if anything fails anywhere it gives up and starts again from the start.

src/What4/ExprHelpers.hs
```diff
@@ -0,0 +1,359 @@
...
+      W4B.BoundVarExpr bv' -> do
+        let idx = N.indexValue (W4B.bvarId bv')
+        stToIO $ H.lookup bvs idx >>= \case
+          Just bvs' -> return $ S.member (Some bv') bvs'
+          _ -> return False
+      _ -> return False
+
+
+isAssert :: W4.SolverSymbol -> Bool
+isAssert symbol = T.isPrefixOf (T.pack "assert") $ W4.solverSymbolAsText symbol
```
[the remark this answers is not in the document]
danmatichuk: Should be good now, the prefix that is used to generate the function names now comes from this module.

src/Pate/Verification.hs
```diff
@@ -891,23 +522,25 @@ getGPValueAndTrace ::
...
     CS.GlobalPair val globs
       | Just mt <- CGS.lookupGlobal mem globs
-      , Just jc <- CGS.lookupGlobal eclass globs
-      , Just rp <- CGS.lookupGlobal rpv globs -> withValid $ do
+      , Just ec <- CGS.lookupGlobal eclass globs -> withValid $ do
         val' <- structToRegState @sym @arch val
-        return $ (val', mt, jc, rp)
-    _ -> throwError undefined
+        return $ (asm, val', mt, ec)
+    _ -> CME.throwError undefined
```
travitch, line 543: Can we turn this into a descriptive error message instead of `undefined`?

PR #36, opened 2021-01-23
Concretereads

danmatichuk:
Support ARM indirect reads by concretizing reads from read-only memory during equivalence checking


src/Pate/SimState.hs
```diff
@@ -115,6 +127,159 @@ simOutRegs ::
 simOutRegs simIn = simRegs $ simOutState simIn
 
 
+newtype ExprSet sym tp where
+  ExprSet :: Set (AsOrd (W4.SymExpr sym) tp) -> ExprSet sym tp
+
+data AsOrd f tp where
+  AsOrd :: OrdF f => { unAsOrd :: f tp } -> AsOrd f tp
```
travitch, line 134: I'm surprised this is necessary - interesting pattern (not opposed)

src/Pate/Verification.hs
```diff
@@ -903,55 +931,6 @@ simplifyPred p = withSym $ \sym -> do
...
-              case res of
-                W4R.Sat _ -> do
-                  return Nothing
-                W4R.Unsat _ -> do
-                  return $ Just e'
-                W4R.Unknown -> do
-                  return Nothing
-          Nothing -> do
-            return Nothing
-      False -> return Nothing
```
travitch, line 943: Nice to see this go

src/Pate/Verification.hs
```diff
@@ -1196,49 +1157,112 @@ validRegister ::
...
+    go mops = do
+      flatops <- fmap F.toList $ liftIO $ MT.flatMemOps sym mops
+      fmap mconcat $ mapM readConcrete flatops
+
+    readConcrete ::
+      MT.MemOp sym (MM.ArchAddrWidth arch) ->
+      EquivM sym arch (AssumptionFrame sym)
+    readConcrete (MT.MemOp (CLM.LLVMPointer reg off) dir _ sz val end) = do
+      case (W4.asNat reg, W4.asBV off, dir) of
+        (Just 0, Just off', MT.Read) -> do
```
travitch, line 1224: Does this mean that we only pull concrete values from region 0?

travitch (approved):

src/Pate/Verification.hs
danmatichuk: Currently, yes. All addresses from the binary are flattened into absolute addresses at region 0, and so this is the only place we would expect to find concrete values.

src/Pate/SimState.hs
danmatichuk: It's just used to create a set of a single type of expression. It felt a bit strange, but it was the most obvious way to connect `Ord` and `OrdF` at the type level for `SymExpr`.

PR #46, opened 2021-02-03
Refactor to use the offline backend

### LINKED ISSUE #37: Replace `rebindExpr` with `rebindWithFrame` to avoid `inNewFrameWithVars`
Proof results (i.e. computed pre-domains in the form of state predicates) are abstracted over the initial state of the machine by representing that state as bound variables. These are rebound to the post-state of the preceding block slice when used as a post-domain target. Currently this rebinding is done using `rebindExpr`, which abuses the internal function definition mechanism of what4 to instantiate those bound variables. Expressions containing bound variables cannot directly be used in solver queries, however, unless they have been explicitly fixed via `What4.Protocol.Online.inNewFrameWithVars`, which adds additional complexity and will make it difficult to transition to offline solvers.

`SimState.rebindWithFrame` is a more general function that supports arbitrary sub-expression rewrites given a binding environment. Using this would allow the initial state to be instead represented as free variables (i.e. `What4.freshConstant`), which would allow us to avoid using `inNewFrameWithVars`. It would have the added benefit of not needing to carry around the evidence that the expressions used to bind the initial state were all necessarily bound variables.

### LINKED ISSUE #39: Switch solving to use the offline backend
This will provide more flexibility over the online backend (and not force us to serialize verification). The transition to parallel verification will be a follow-on effort.

This depends on #37

travitch:
This is the essence of the change, though there are still some unbound variables that get sent to the solver that will need to be addressed.

Fixes #37 
Fixes #39

### DIFF
--- a/src/Pate/CounterExample.hs
+++ b/src/Pate/CounterExample.hs
@@ -161,6 +161,7 @@ groundBlockEndCase fn blkend = withSym $ \sym -> do
   groundMuxTree fn blkend_tree
 
 groundTraceDiff :: forall sym arch.
+  HasCallStack =>
   SymGroundEvalFn sym ->
   EquivRelation sym arch ->
   SimBundle sym arch ->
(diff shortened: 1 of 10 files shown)


travitch:
Very cool changes!

danmatichuk (approved):

PR #72, opened 2021-03-31
Update to the latest crucible/what4

travitch:
This deals with the demotion of Nat from a base type and includes corresponding
updates to semmc/asl-translator/macaw and others.

### DIFF
--- a/src/Pate/Equivalence.hs
+++ b/src/Pate/Equivalence.hs
@@ -389,13 +389,13 @@ stateEquivalence ::
   PA.ValidArch arch =>
   sym ->
   -- | stack memory region
-  W4.SymExpr sym W4.BaseNatType ->
+  W4.SymNat sym ->
   EquivRelation sym arch
 stateEquivalence sym stackRegion =
   let
     isStackCell cell = do
       let CLM.LLVMPointer region _ = PMC.cellPtr cell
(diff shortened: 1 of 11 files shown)


travitch:
@danmatichuk This could probably use another pair of eyes. The tests pass, but I'm a bit nervous about the nat->int and int->nat conversions. I think it should always be fine, except possibly for models coming back from the solver. I just map any negative values to the positive range, but perhaps that could generate a false conflict? We could potentially assert some side conditions when asking for models.

travitch:
Actually, I think we are getting a safe bound - I'm allocating our symbolic block IDs using `freshNat`, which properly constrains the underlying symbolic integer.

travitch:
Ah okay, here is a question for @danmatichuk - do we ever make a fresh LLVMPointer struct (in the what4 base types struct sense)? Or do we ever have an uninterpreted function that magically introduces one? If so, we would probably need to project out the (now Integer) block ID field and constrain it

danmatichuk:
The only place I can think of is the result of pointer operations, when undefined, is the result of an uninterpreted function which yields an LLVPointer struct.

https://github.com/GaloisInc/pate/blob/825893b5c3e87d7da8663b4c95d95e038cb11702/src/Pate/Memory/MemTrace.hs#L212

PR #75, opened 2021-04-08
Test out self-hosted runners

jared-w:
Let's give this a whirl eh?


travitch:
@jared-w I tested this a little bit - the runner(s?) seemed to pick up one of my jobs but left the rest hanging for some reason

jared-w:
While testing this I briefly played with adopting a `cabal.project` instead of submodules and ran into a few interesting bugs that way. The tl;dr is that a few cabal files are wrong across the binary analysis tools (meaning that they don't work if the package is from cabal sdist vs just source-code locally).

Another interesting bug was that some code uses cabal's data file support to load bundled files at compile time with template haskell. It turns out that this is always invalid and only accidentally happens to work if your code is a submodule. This is the same situation that github's semantic codebase [ran into](https://github.com/github/semantic/issues/622) for the same exact reason. So, I'll open up an issue in the right repository, but I'm going to also suggest AOT code gen to embed the files directly into the haskell code.

---

That said, this is ready to go and is quite stable now. I'm not sure how it'll hold up to being used "at scale" but it's worth experimenting with now, I think.

PR #153, opened 2021-11-19
RWD refactorings

robdockins:
A variety of fixes and refactorings.

The main difference is a significant simplification of the `ParsedFunctionMap` datastructure.  It is now simply a mapping from function entry points to blocks contained in that function. The interval map structure previously was used mostly so that we could compute all the blocks forward of an address.  We no longer need to do this.

`ConcreteBlock` now contains only only the head of a block, but also the function entry point for the function being considered. This means we no longer have to guess what function the block is part of.

Moreover, we compute a proper depth-first search on blocks in the symbolic execution phase to precisely find back edges in the control flow graphs instead of relying on heuristics.


src/Pate/Discovery.hs
```diff
@@ -309,58 +301,81 @@ concreteValueAddress ::
...
-    blk_f <- mkConcreteBlock PB.BlockEntryJump f
-    return $ [ BlockTarget blk_t Nothing, BlockTarget blk_f Nothing ]
-  MD.ParsedLookupTable _jt st _ _ -> go (concreteNextIPs st) Nothing
+  EquivM sym arch [PB.BlockTarget arch bin]
+concreteJumpTargets from pb = case MD.pblockTermStmt pb of
+  MD.ParsedCall st ret ->
+    callTargets from (concreteNextIPs st) ret PB.BlockEntryPostFunction
+
+  MD.PLTStub st _ _ ->
+    case MapF.lookup (MC.ip_reg @(MC.ArchReg arch)) st of
```
travitch, line 328: For `RegState`, you can use the `boundValue` lens (https://github.com/GaloisInc/macaw/blob/master/base/src/Data/Macaw/CFG/Core.hs#L589). `st ^. boundValue (ip_reg @arch)`.

This is conceptually total (our guarantees of this are well-intentioned and usually realized)

travitch (commented):
Thanks, big fan of the simplifications

src/Pate/Discovery.hs
robdockins: Is there a particular reason PLTStub has a `MapF (ArchReg arch) (Value arch ids)` instead of a `RegState (ArchReg arch) (Value arch ids)`?

travitch: Damn, I didn't realize that wasn't a `RegState`. It looks like the `PLTStub` version is actually a sparse map.

This case might just not make sense (and I'd be surprised if we are actually executing these stubs). Even if you found a PC value there, it would probably be to a computed value in a shared library. It is not the address returned to. Unfortunately, `PLTStub`s can't know where they are returning to.

During execution, the program issues a call to the PLT stub, putting the return address either on the stack or into a link register. The PLT stub then tail calls into a shared library, which will issue the return.

robdockins: I'm not sure what that means for this function.  Should we change this case?

travitch: My hypothesis is that this case should just be a panic - I don't think we ever reach it. If we do reach it, we don't have a way to handle it properly. The way it should be is that, if we see a call to a shared library function through a PLT stub, we just make an "external call" frame using that `ExternalDomain` type. If we aren't doing that right now, that is a separate bug.

robdockins: OK, I'll make this an error and see what happens to the test suite.

robdockins: Well, changing this breaks several test cases in the ppc test suite. I don't really know what that means.

travitch: Which tests? PPC has not been thoroughly tested yet, so that may not be a huge problem

travitch:
I thought some more about the PPC failure. I think it means that macaw just completely fails to identify PLT stubs on ARM. Instead, they probably just look like tail calls embedded in the caller.

We probably want it to be the case that, if we see a call or jump to a PLT stub block, we just pretend it doesn't exist and use the external domain

PR #156, opened 2021-12-03
Tr/incremental code discovery

### LINKED ISSUE #155: Make code discovery incremental
Currently, we run code discovery via macaw at the beginning of the verification phase. This discovers all of the code in a binary, which allows us to quickly look up each function.

This is problematic for very large binaries (particularly those that link in large numbers of static libraries), as the initial code discovery can consume all memory. We should modify the verifier to only decode functions on demand (and possibly maintain a LRU cache).

travitch:
Previously, we used macaw to eagerly run code discovery and analyze all of the
code that we could find. This was problematic for large binaries, for which we
ran out of memory. The memory expansion of macaw IR (and Crucible) compared to
machine code is enormous (probably over 1000x).

With this change, we now only run code discovery on demand (i.e., when we find a
call to a function, we analyze just that function).  We cache analyzed
functions for easy access.  Note that cache currently only grows; it could
trivially become an LRU cache if desired.

An unfortunate side effect of this change is that we can do a bit less in the
way of diagnostics. Before, we were able to figure out which hints from users
were incorrect (i.e., inconsistent with analysis results). Now that we only
lazily find code, we can't do that anymore. Some of our diagnostics and metrics
are less precise now.

Closes #155

### DIFF
--- a/src/Pate/Discovery.hs
+++ b/src/Pate/Discovery.hs
@@ -29,6 +29,7 @@ import           Control.Monad.IO.Class ( liftIO )
 import qualified Control.Monad.Reader as CMR
 import qualified Data.BitVector.Sized as BVS
 import qualified Data.Foldable as F
+import qualified Data.List.NonEmpty as DLN
 import qualified Data.Map.Strict as Map
 import           Data.Maybe ( catMaybes )
 import qualified Data.Parameterized.Classes as PC
@@ -418,9 +419,9 @@ runDiscovery ::
   CME.ExceptT (PEE.EquivalenceError arch) IO ([Word64], PMC.BinaryContext arch bin)
 runDiscovery mCFGDir repr elf hints = do
   let archInfo = PLE.archInfo elf
-  entries <- F.toList <$> MBL.entryPoints bin
+  entries <- MBL.entryPoints bin
 
-  let (invalidHints, hintedEntries) = F.foldr (addFunctionEntryHints (Proxy @arch) mem) ([], entries) (PH.functionEntries hints)
+  let (invalidHints, hintedEntries) = F.foldr (addFunctionEntryHints (Proxy @arch) mem) ([], F.toList entries) (PH.functionEntries hints)
   let ds = MD.cfgFromAddrs archInfo mem Map.empty hintedEntries []
 
   -- If the user wants to persist macaw CFGs, do so here
@@ -438,7 +439,7 @@ runDiscovery mCFGDir repr elf hints = do
   let idx = F.foldl' addFunctionEntryHint Map.empty (PH.functionEntries hints)
   let fnmap = PMC.buildFunctionEntryMap repr (ds ^. MD.funInfo)
 
-  let startEntry = head entries -- can't fail because entryPoints returns a NonEmpty
+  let startEntry = DLN.head entries
   startEntry' <- lookupFunctionEntry fnmap (PA.segOffToAddr startEntry)
 
   abortFnEntry <- traverse
--- a/src/Pate/Discovery.hs
+++ b/src/Pate/Discovery.hs
@@ -25,7 +25,7 @@ import           Control.Lens ( (^.) )
 import           Control.Monad (forM)
 import qualified Control.Monad.Catch as CMC
 import qualified Control.Monad.Except as CME
-import           Control.Monad.IO.Class ( liftIO )
+import           Control.Monad.IO.Class ( liftIO, MonadIO )
 import qualified Control.Monad.Reader as CMR
 import qualified Data.BitVector.Sized as BVS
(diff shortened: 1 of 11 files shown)


src/Pate/Discovery.hs
```diff
@@ -407,6 +413,18 @@ lookupFunctionEntry fnmap a =
...
+  :: (w ~ MC.ArchAddrWidth arch, MM.MemWidth w)
+  => MM.Memory w
+  -> MD.AddrSymMap w
+  -> PH.FunctionDescriptor
+  -> CME.ExceptT (PEE.EquivalenceError arch) IO (MD.AddrSymMap w)
+addAddrSym mem m funcDesc = do
+  let symbol = TE.encodeUtf8 (PH.functionSymbol funcDesc)
+  case PM.resolveAbsoluteAddress mem (fromIntegral (PH.functionAddress funcDesc)) of
+    Just segoff -> return (Map.insert segoff symbol m)
+    Nothing -> error "Invalid function entry point"
```
travitch, line 426: Make this into a panic

src/Pate/Discovery.hs
```diff
@@ -418,34 +436,27 @@ runDiscovery ::
...
-  -- If the user wants to persist macaw CFGs, do so here
-  --
-  -- Save these CFGs to <dir>/<repr>/<address>
-  liftIO $ F.forM_ mCFGDir $ \cfgDir -> do
-    let baseDir = cfgDir </> show repr
-    SD.createDirectoryIfMissing True baseDir
-    F.forM_ (ds ^. MD.funInfo) $ \(Some dfi) -> do
-      let fname = baseDir </> show (MD.discoveredFunAddr dfi) <.> "cfg"
-      IO.withFile fname IO.WriteMode $ \hdl -> do
-        PPT.hPutDoc hdl (PP.pretty dfi)
```
travitch, line 435: Need to restore this (probably at the incremental discovery stage - write out CFGs as we find them)

src/Pate/Verification.hs
```diff
@@ -253,15 +249,26 @@ verifyPairs validArch@(PA.SomeValidArch _ _ hdr) logAction elf elf' vcfg pd = do
...
   PA.ValidArch arch =>
   PMC.BinaryContext arch bin ->
   PC.BlockData ->
   CME.ExceptT (PEE.EquivalenceError arch) IO (PB.FunctionEntry arch bin)
 unpackBlockData ctxt (PC.Hex w) =
-  PD.lookupFunctionEntry (PMC.functionEntryMap ctxt) (PAd.memAddrToAddr (MM.absoluteAddr (MM.memWord w)))
+  case PM.resolveAbsoluteAddress mem (fromIntegral w) of
+    Just segAddr ->
+      return PB.FunctionEntry { PB.functionSegAddr = segAddr
+                              , PB.functionSymbol = undefined
```
travitch, line 262: Fill this in

src/Pate/Monad/Context.hs
```diff
@@ -92,13 +94,80 @@ funInfoToFunEntry binRepr dfi =
...
+  let mem = MD.memory ds1
+  -- First, check if we have a cached set of blocks for this state
+  case Map.lookup (PB.blockFunctionEntry blk) (parsedFunctionCache st) of
+    Just blks -> return (Just blks)
+    Nothing -> do
+      -- Otherwise, run code discovery at this address
+      case MM.resolveRegionOff mem (MM.addrBase baddr) (MM.addrOffset baddr) of
+        Nothing -> return Nothing -- This could be an exception... but perhaps better as a proof failure node
+        Just faddr -> do
+          (Some dfi, blks) <- IORef.atomicModifyIORef' pfmRef (atomicAnalysis faddr)
```
robdockins, line 149: Do you actually want the strict atomic modify here?  It might be better to allow it to be lazy, as the code discovery process itself is pure.  The lazy version might get you less contention on this IORef.

robdockins: Or... maybe it doesn't matter much, given that the discovery state has to be threaded through.

robdockins (approved):

src/Pate/Monad/Context.hs
travitch: I spent some time trying to reason through what would actually happen here (e.g., if there are retry semantics, and what they are). I'm at "I'm not sure". The returned value is a pair and the value stored in the IORef doesn't have strict fields, so maybe I should just make it lazy to make it clearer that it isn't trying hard to force things to normal form.

travitch: I will add a note before merging, since I agree that it is unclear and something that might need to be addressed if there ends up being some weird contention that leads to a lot of re-computation

PR #174, opened 2021-12-24
Refactor the memory model in preparation for tracking block IDs

travitch:
The main goals of the refactor are to:

1. Separate the core memory model data structure from the implementation of
primitive crucible support operations (only the former need to be augmented for
the memory model improvements)
2. Remove unnecessary duplication (e.g., the symbolic sequence has a shared
implementation in crucible)
3. Clarify semantics
4. Document interfaces

Note that this is a work in progress


thebendavis:
out of date / subsumed

PR #179, opened 2022-01-12
Add abstraction for memory model representation

### LINKED ISSUE #136: Abstract low-level memory equivalence checks in the memory model
Currently we break the memory model abstraction in [memEqAtRegion](https://github.com/GaloisInc/pate/blob/6e3788f716ad7d9c2ff1dc99d385826a0f90a8fb/src/Pate/Equivalence.hs#L469)
and [memEqOutsideRegion](https://github.com/GaloisInc/pate/blob/6e3788f716ad7d9c2ff1dc99d385826a0f90a8fb/src/Pate/Equivalence.hs#L455), which assert total memory equality either inclusive or exclusive of a given region.

To account for potential chances to the underlying memory representation, these checks should be pushed behind the abstraction boundary of the memory model.

### LINKED ISSUE #175: Tracking memory model abstraction violations
There are a number of places in the verifier that violate what should be abstraction boundaries in the memory model. These significantly complicate any progress on improving the memory model (either for performance or soundness). This issue is intended to track the violations we identify, determine what semantics are required to support them, and transform them into structured APIs presented by the memory model for consumption in the rest of the verifier. Some of these are taken from #97 (reported by @dmwit)

- [ ] `bindMemory` in Pate.Verification.Domain
- [ ] `flatVarBinds` in Pate.SimState
- [ ] `memPredPre` in Pate.Equivalence
- [ ] `guessMemoryDomain` in Pate.Verification.Domain (caller of `bindMemory`)
- [ ] `groundBV` in Pate.Proof.Ground (calls `getPointerTags`, which traverses formulas to find undefined pointer operations)

Around several of these, there are additional direct accesses to the `memArr` of the memory model implementation. All of these need to be eliminated and abstracted into a structured interface to keep the memory model abstract.

danmatichuk:
The verifier now avoids directly using the underlying array
from the memory model

Fixes #136 
Fixes #175

### DIFF
--- a/src/Pate/Memory/MemTrace.hs
+++ b/src/Pate/Memory/MemTrace.hs
@@ -28,11 +28,11 @@
 
 module Pate.Memory.MemTrace
 ( MemTraceImpl(..)
-, MemTraceVar(..)
+, MemTraceState
 , MemTrace
 , llvmPtrEq
-, readMemArr
-, writeMemArr
+, readMemState
+, writeMemState
 , MemFootprint(..)
 , MemOpDirection(..)
 , getCond
@@ -41,7 +41,6 @@ module Pate.Memory.MemTrace
 , UndefPtrOpTag
 , UndefPtrOpTags
 , UndefinedPtrOps(..)
-, initMemTraceVar
 , MemOpCondition(..)
 , MemOp(..)
 , flatMemOps
@@ -49,8 +48,12 @@ module Pate.Memory.MemTrace
 , initMemTrace
 , classifyExpr
 , mkMemTraceVar
+, mkMemoryBinding
 , mkUndefinedPtrOps
 , macawTraceExtensions
+, memEqOutsideRegion
+, memEqAtRegion
+, memEqExact
 ) where
 
 import Unsafe.Coerce
@@ -73,7 +76,7 @@ import qualified Data.Parameterized.Context as Ctx
 
 import qualified Data.Macaw.Types as MT
 import Data.Macaw.CFG.AssignRhs (ArchAddrWidth, MemRepr(..))
-import Data.Macaw.Memory (AddrWidthRepr(..), Endianness(..), MemWidth, addrWidthClass, addrWidthNatRepr)
+import Data.Macaw.Memory (AddrWidthRepr(..), Endianness(..), MemWidth, addrWidthClass, addrWidthRepr, addrWidthNatRepr)
 import Data.Macaw.Symbolic.Backend (MacawEvalStmtFunc, MacawArchEvalFn(..))
 import Data.Macaw.Symbolic ( MacawStmtExtension(..), MacawExprExtension(..), MacawExt
                            , GlobalMap, MacawSimulatorState(..)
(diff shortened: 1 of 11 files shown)


src/Pate/Memory/MemTrace.hs
```diff
@@ -504,11 +507,12 @@ instance TestEquality (SymExpr sym) => Eq (MemOp sym ptrW) where
...
   { memSeq :: MemTraceSeq sym ptrW
   -- ^ the sequence of memory operations in execution order
-  , memArr :: MemTraceArr sym ptrW
+  , memState :: MemTraceState sym ptrW
   -- ^ the logical contents of memory
   }
 
-data MemTraceVar sym ptrW = MemTraceVar (SymExpr sym (MemArrBaseType ptrW))
+data MemTraceState sym ptrW = MemTraceState
+  { memArr :: MemTraceArr sym ptrW }
```
travitch, line 515: Seems like we can remove `MemArrBaseType` now

src/Pate/Memory/MemTrace.hs
```diff
@@ -540,23 +544,24 @@ initMemTrace ::
...
-  AddrWidthRepr ptrW ->
-  IO (MemTraceImpl sym ptrW, MemTraceVar sym ptrW)
-initMemTraceVar sym Addr32 = do
-  arr <- ioFreshConstant sym "InitMem" knownRepr
-  return $ (MemTraceImpl mempty arr, MemTraceVar arr)
-initMemTraceVar sym Addr64 = do
-  arr <- ioFreshConstant sym "InitMem" knownRepr
-  return $ (MemTraceImpl mempty arr, MemTraceVar arr)
+  MemTraceState sym ptrW ->
+  -- ^ initial memory state (appears in the the given expression when the binding is applied)
```
travitch, line 556: If you want to put the arrows at the end of the line, you probably want to introduce this comment before the argument using `-- |`

src/Pate/Panic.hs
```diff
@@ -1,10 +1,12 @@
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE PackageImports #-}
+
 module Pate.Panic (
   P.panic,
   PateComponent(..)
   ) where
 
-import qualified Panic as P
+import qualified "panic" Panic as P
```
travitch, line 9: What happened here?

src/What4/ExprHelpers.hs
```diff
@@ -181,23 +190,139 @@ anyPred ::
...
+          if (W4B.nonceExprApp a0) == a0' then return e
+          else IO.liftIO $ W4B.sbNonceExpr sym a0'
+        _ -> return e
+  (e', binds) <- CMW.runWriterT (go e_outer)
+  Pair vars vals <- return $ toBindings binds
+  fn <- W4.definedFn sym W4.emptySymbol vars e' W4.AlwaysUnfold
+  W4.applySymFn sym fn vals >>= fixMux sym
+
+
+type ExprBindings sym = MapF.MapF (W4.SymExpr sym) (W4.SymExpr sym)
```
travitch, line 291: Can we just say a bit here about what is in the map? Are there any interesting invariants here?

travitch (commented):
Thanks, this is a really helpful change

PR #189, opened 2022-02-16
Swap out the `MVar`-based futures implementation for

robdockins:
one based on the `async` library.

This basically seems to work, but does not fix the nondeterministic
test suite failures as I hoped it would. Not clear if it is worth
doing on it's own.


travitch:
In principle I really like relying on someone else's careful thinking about exception safety

robdockins:
The SP verifier doesn't make use of the lazy proof combinators, so I'm not sure it makes sense to spend much effort in this direction.

PR #194, opened 2022-03-03
Refactor proof structure and counter-example grounding

danmatichuk:
This simplifies the proof structure by removing the 'prf' type parameter in favor of more explicitly using 'sym' and 'arch'. Additionally this introduces 'Pate.Ground' as a general mechanism for capturing a model from the solver with respect to the inner expressions of some expression-containing type, which is now used to represent the counter-examples from the verifier.


src/Pate/Proof.hs
```diff
@@ -143,82 +136,100 @@ instance PEM.ExprMappable sym (EquivTripleBody sym arch) where
...
 type ProofDomainType = 'ProofDomainType
 type ProofTripleType = 'ProofTripleType
 type ProofFunctionCallType = 'ProofFunctionCallType
 type ProofBlockSliceType = 'ProofBlockSliceType
 type ProofStatusType = 'ProofStatusType
-type ProofInlineType = 'ProofInlineType
 
 
-data VerificationStatus ce =
+data InequivalenceResultC arch grnd where
```
travitch, line 153: Is there something that the `C` suffix is priming me for?

src/Pate/Proof.hs
```diff
@@ -143,82 +136,100 @@ instance PEM.ExprMappable sym (EquivTripleBody sym arch) where
...
+instance PEM.ExprMappable sym (VerificationStatus sym arch) where
+  mapExpr sym f s = case s of
+    VerificationFail ce condeq -> do
+      ce' <- PEM.mapExpr sym f ce
+      condeq' <- PEM.mapExpr sym f condeq
+      return $ VerificationFail ce' condeq'
+    _ -> return s
 
 -- | An abstract proof object, representing the overall structure of an equivalence proof.
 -- The type parameter 'prf' abstracts this structure over the specific types used in the proof,
```
travitch, line 228: I think this comment might need to be updated now that the type parameters have changed

src/Pate/Proof.hs
```diff
@@ -276,73 +287,87 @@ data ProofApp prf (node :: ProofNodeType -> DK.Type) (tp :: ProofNodeType) where
...
+     MemoryDomain
+       <$> (MapF.fromList <$> (traverse transMemCell (MapF.toList a1)))
+       <*> f a2
+   where
+     transMemCell (MapF.Pair cell (Const p)) = MapF.Pair
+       <$> PEM.mapExpr sym f cell
+       <*> (Const <$> f p)
 
 -- | Traverse the nodes of a 'ProofApp', changing the
 -- recursive 'node' type while leaving the leafs (i.e. the 'prf' type) unchanged.
```
travitch, line 355: This comment also references the removed `prf` type parameter

src/Pate/Proof.hs
```diff
@@ -353,296 +378,162 @@ traverseProofApp f = \case
...
-  ProofDomain a1 a2 a3 a4 -> ProofDomain
-    <$> pure a1
-    <*> pure a2
-    <*> pure a3
-    <*> pure a4
+  ProofDomain a1 -> ProofDomain <$> pure a1
+  ProofStatus a1 -> ProofStatus <$> pure a1
 
 -- | Map over the nodes of a 'ProofApp', changing the
 -- 'node' type while leaving the leafs (i.e. the 'prf' type) unchanged.
```
travitch, line 385: `prf` comment again

src/Pate/Proof.hs
```diff
@@ -353,296 +378,162 @@ traverseProofApp f = \case
...
-  Monad m =>
-  IsBoolLike prf m =>
-  m (ProofMemoryDomain prf)
-emptyMemDomain = do
-  falsePred <- proofPredFalse @prf
-  return $ ProofMemoryDomain MapF.empty falsePred
-
-type family ProofScope prf :: DK.Type
+type family SymScope sym :: DK.Type
+type instance SymScope (W4B.ExprBuilder t fs scope) = t
```
travitch, line 421: It is a bit hard to eyeball where this is used. On one hand it is kind of funny that `SymScope` doesn't extract the type parameter named `scope`.

If this is only used in a few places (e.g., `ProofNonce` is the only one I saw). If that really is the only use, could we avoid these potentially confusing types by just defining
```
data ProofNonce sym (tp :: ProofNodeType) where
  ProofNonce :: (sym ~ W4B.ExprBuilder t fs scope) => N.Nonce t -> ProofNonce sym tp
```

travitch: If it is used widely and is important, fair enough...

src/Pate/Proof.hs
```diff
@@ -353,296 +378,162 @@ traverseProofApp f = \case
...
   testEquality (ProofNonce n1) (ProofNonce n2) = testEquality n1 n2
 
 instance OrdF (ProofNonce prf) where
   compareF (ProofNonce n1) (ProofNonce n2) = compareF n1 n2
 
-proofNonceValue :: ProofNonce prf tp -> Natural
+proofNonceValue :: ProofNonce sym tp -> Natural
 proofNonceValue (ProofNonce n) = fromIntegral (N.indexValue n)
 
 -- | A proof expression, annotated with nonces.
```
travitch, line 441: It isn't just that it adds nonces - the `ProofNonceExpr` encodes the tree structure of the proof with its parent pointers

src/Pate/Proof.hs
```diff
@@ -353,296 +378,162 @@ traverseProofApp f = \case
...
-type ProofNonceApp prf tp = ProofApp prf (ProofNonceExpr prf) tp
+type ProofNonceApp sym arch tp = ProofApp sym arch (ProofNonceExpr sym arch) tp
 
 -- | Strip the nonces from a 'ProofNonceExpr', yielding an equivalent 'ProofExpr'.
-unNonceProof :: ProofNonceExpr prf tp -> ProofExpr prf tp
+unNonceProof :: ProofNonceExpr sym arch tp -> ProofExpr sym arch tp
 unNonceProof prf = ProofExpr $ mapProofApp unNonceProof (prfNonceBody prf)
 
 -- | The semantics of executing two block slices: an original and a patched variant.
 -- The type parameter 'prf' abstracts this over the specific types,
```
travitch, line 456: Another `prf` type parameter reference

src/Pate/Proof/CounterExample.hs
```diff
@@ -93,24 +94,45 @@ import qualified What4.PathCondition as WPC
...
+  EquivM sym arch (PF.InequivalenceResult arch)
 getInequivalenceResult defaultReason pre post slice fn = do
-  groundPre <- groundProofExpr fn pre
-  groundPost <- groundProofExpr fn post
-  gslice <- groundSlice fn slice
-  let reason = fromMaybe defaultReason (getInequivalenceReason groundPost (PF.slBlockPostState gslice))
-  return $ PFI.InequivalenceResult gslice groundPre groundPost reason
-
+  result <- ground fn $ PF.InequivalenceResult slice pre post defaultReason
+  TF.forF result $ \result' -> PG.withGroundSym result $ \groundResult -> do
```
travitch, line 111: Given that this is an unfamiliar traversable instance, it would be worth a note here about what the semantics are intended to be here. Is this traversing all of the sub-components? What should I expect that `result'` is here? I was a bit surprised that `result` is being passed to `withGroundSym` rather than `result'`, so I'm not sure I understand what this is supposed to be doing. If that is the case, can the `withGroundSym` be outside of the `forF`?

src/Pate/Proof/CounterExample.hs
```diff
@@ -182,27 +204,30 @@ getPathCondition ::
...
   EquivM sym arch (PPa.PatchPairC (W4.Pred sym))
 getPathCondition bundle slice dom fn = withSym $ \sym -> do
-  groundDom <- groundProofExpr fn dom
   let
+    regInDomain ::
+      MM.ArchReg arch tp -> EquivM sym arch Bool
+    regInDomain reg =
+      execGroundFn fn $ getConst $ PF.eqDomainRegisters dom ^. MM.boundValue reg
+
     getRegPath ::
```
travitch, line 218: This might be something to add in a different PR (perhaps in the upcoming cleanups), but what is the intent in the logic here with `regInDomain`? What is the intent of `getRegPath` (especially in the case where `not (regInDomain)`?

src/Pate/Proof/Instances.hs
```diff
@@ -102,135 +92,48 @@ import qualified What4.Interface as W4
 import qualified What4.ExprHelpers as WEH
 import qualified What4.Expr.Builder as W4B
 import qualified What4.Expr.GroundEval as W4G
+import qualified What4.Partial as W4P
 
 data SomeProofSym (arch :: DK.Type) tp where
```
travitch, line 97: How that there is no more `ProofSymNonceExpr`, is this maybe more of a `SomeProofNonceExpr`? Or is that name taken already?

src/Pate/Proof/Instances.hs
```diff
@@ -102,135 +92,48 @@ import qualified What4.Interface as W4
...
+-- type instance PF.ProofPredicate (ProofSym sym arch) = W4.Pred sym
+-- type instance PF.ProofCounterExample (ProofSym sym arch) = InequivalenceResult arch
+-- type instance PF.ProofCondition (ProofSym sym arch) = CondEquivalenceResult sym arch
+-- type instance PF.ProofMacawValue (ProofSym sym arch) = PSR.MacawRegEntry sym
+-- type instance PF.ProofBlockExit (ProofSym sym arch) = CS.RegValue sym (MS.MacawBlockEndType arch)
+-- type instance PF.ProofContext (ProofSym sym arch) = PPa.PatchPair (PS.SimState sym arch)
+-- type instance PF.ProofScope (ProofSym (W4B.ExprBuilder t st fs) arch) = t
+-- type instance PF.ProofInlinedResult (ProofSym sym arch) = PVM.SomeWriteSummary (MM.ArchAddrWidth arch)
+-- 
+-- type SymDomain sym arch = PF.ProofExpr (ProofSym sym arch) PF.ProofDomainType
```
travitch, line 125: Can probably delete this too

src/Pate/Proof/Instances.hs
```diff
@@ -102,135 +92,48 @@ import qualified What4.Interface as W4
...
-type ProofSymApp sym arch = PF.ProofApp (ProofSym sym arch) (ProofSymExpr sym arch)
+
+-- instance (PA.ValidArch arch, PSo.ValidSym sym) => PEM.ExprMappable sym (PF.ProofExpr sym arch tp) where
+--   mapExpr sym f (PF.ProofExpr app) = PF.ProofExpr <$> PEM.mapExpr sym f app
+-- 
+-- instance (PA.ValidArch arch, PSo.ValidSym sym) => PEM.ExprMappable sym (PF.ProofNonceExpr sym arch tp) where
+--   mapExpr sym f (PF.ProofNonceExpr nonce parent app) = do
+--     app' <- PEM.mapExpr sym f app
+--     return $ PF.ProofNonceExpr nonce parent app'
+--
```
travitch, line 136: And this

src/Pate/Proof/Instances.hs
```diff
@@ -438,22 +352,23 @@ collapseProofMemoryDomain memdom = mapMaybe go $ MapF.toList (PF.prfMemoryDomain
...
+-- type instance PF.ProofCounterExample (ProofGround arch) = InequivalenceResult arch
+-- type instance PF.ProofCondition (ProofGround arch) = ()
+-- type instance PF.ProofPredicate (ProofGround arch) = Bool
+-- type instance PF.ProofMacawValue (ProofGround arch) = GroundMacawValue
+-- type instance PF.ProofBlockExit (ProofGround arch) = GroundBlockExit arch
+-- -- no additional context needed for ground values
+-- type instance PF.ProofContext (ProofGround arch) = ()
+-- type instance PF.ProofInlinedResult (ProofGround arch) = PVM.SomeWriteSummary (MM.ArchAddrWidth arch)
+-- 
+-- instance PA.ValidArch arch => PF.IsProof (ProofGround arch)
```
travitch, line 371: And these

src/Pate/Verification.hs
```diff
@@ -509,9 +509,9 @@ checkEquivalence triple = startTimer $ withSym $ \sym -> do
 withSimBundle ::
   (HasCallStack, PEM.ExprMappable sym f) =>
   PPa.BlockPair arch ->
-  (SimBundle sym arch -> EquivM sym arch f) ->
-  EquivM sym arch (SimSpec sym arch f)
-withSimBundle pPair f = withEmptyAssumptionFrame $ withSym $ \sym -> do
+  (SimBundle sym arch -> EquivM sym arch (f, prf)) ->
+  EquivM sym arch (SimSpec sym arch f, prf)
```
travitch, line 513: How should I interpret this return value? What is the concept I should have in mind when calling `withSimBundle`?

travitch (changes requested):
I like this simplification a lot. Most of my comments are about documentation updates/requests

src/Pate/Proof.hs
```diff
@@ -353,296 +388,331 @@ traverseProofApp f = \case
...
+--     <*> pure a4
+--     
+--   ProofStatus a1 -> ProofStatus
+--     <$> traverse (\(cex, cond) -> (,) <$> prfCounterExampleTrans f cex <*> prfConditionTrans f cond) a1
+--   ProofDomain a1 a2 a3 a4 -> ProofDomain
+--     <$> MM.traverseRegsWith (\_ (Const v) -> Const <$> prfPredTrans f v) a1
+--     <*> transformMemDomain f a2
+--     <*> transformMemDomain f a3
+--     <*> prfContextTrans f a4
+--
```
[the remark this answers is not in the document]
danmatichuk: Done

src/What4/ExprHelpers.hs
```diff
@@ -108,6 +109,13 @@ import qualified What4.Symbol as WS
 import           Data.Parameterized.SetF (SetF)
 import qualified Data.Parameterized.SetF as SetF
 
+
+-- FIXME: this is a workaround for the fact that 'natToInteger' from What4.Interface
+-- is in the IO monad, but doesn't need to be.
+natToIntegerPure ::
+  W4.IsExprBuilder sym => sym -> W4.SymNat sym -> W4.SymInteger sym
+natToIntegerPure _sym n = unsafeCoerce n
```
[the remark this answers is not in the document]
danmatichuk: I was hoping for a change in the What4 interface, since this function really does not need to be in IO. I raised an issue here: https://github.com/GaloisInc/what4/issues/193.

Plumbing through the necessary IO to get this working is non-trivial, since this is at the bottom of basically anything that looks at ground values. 

An alternative fix is to address this issue here: https://github.com/GaloisInc/pate/issues/195 by making `integerToNat` more stable (such that unwrapping and wrapping a SymNat does not change the inner term at all).

src/Pate/Proof/CounterExample.hs
danmatichuk: This was originally due to some indecision about how `Grounded` values should be allowed to be modified. I cleaned this up to just have a single `traverseWithGroundSym` which makes this look more like you would expect.

src/Pate/Proof/CounterExample.hs
danmatichuk: This just avoids examining the terms for registers which are outside of the equivalence domain (i.e. they were not part of the goal equivalence proof and therefore should not be examined when trying to come up with a path condition)

src/Pate/Verification.hs
danmatichuk: Added some documentation

src/Pate/Proof/Instances.hs
danmatichuk: Yeah this makes sense

src/Pate/Proof.hs
danmatichuk: It's used in at least one other place, which is to supply the type parameter to the nonce generator in `Pate.Monad.Environment`. It's not obvious to me how to write that easily without the type family.

src/Pate/Proof.hs
danmatichuk: Expanded on this comment a bit

travitch (approved):
Thanks - the additional clarifications are useful

src/What4/ExprHelpers.hs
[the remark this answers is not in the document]
robdockins: I think the necessary What4 change should just require a few lines to export a new, pure, extraction function.  I don't see any problem with widening the API this way.  We already have a What4 submodule, so no need to wait for a release.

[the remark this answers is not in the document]
travitch: Thanks. I just wanted to make sure there wasn't some future API change you had in mind that would be in conflict

PR #201, opened 2022-03-07
Simplify 'MemPred' representation and identify it with Proof.MemoryDomain

danmatichuk:
This flattens the structure of 'MemPred' (renaming it to MemoryDomain) and identifies it with the Proof.MemoryDomain type.


src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -0,0 +1,224 @@
...
+-- | Trivial domain that covers all of memory
+universal :: W4.IsExprBuilder sym => sym -> MemoryDomain sym arch
+universal sym = MemoryDomain M.empty (W4.falsePred sym)
+
+-- | Trivial domain that covers no memory (the empty domain)
+empty :: W4.IsExprBuilder sym => sym -> MemoryDomain sym arch
+empty sym = MemoryDomain M.empty (W4.truePred sym)
+
+
+fromFootPrints ::
```
travitch, line 179: What is the concept for this function? There is some non-trivial logic around memory operation directions that probably merits discussion.

What expectations should the reader have for the resulting `MemoryDomain`?

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -0,0 +1,224 @@
...
+      MT.Read -> return $ W4.truePred sym
+      MT.Write -> return $ W4.falsePred sym
+    polarityMatches <- W4.isEq sym polarity dirPolarity
+    cond' <- W4.andPred sym polarityMatches (MT.getCond sym cond)
+    case W4.asConstantPred cond' of
+      Just False -> return Nothing
+      _ -> return $ Just (Some (PMC.MemCell ptr w end), cond')
+  fromList sym locs polarity
+
+addFootPrints ::
```
travitch, line 198: To an existing `MemoryDomain`?

src/Pate/Equivalence/StatePred.hs
```diff
@@ -61,12 +61,12 @@ muxStatePred sym p predT predF = case WI.asConstantPred p of
       (M.zipWithAMatched (\_ p1 p2 -> WI.baseTypeIte sym p p1 p2))
       (predRegs predT)
       (predRegs predF)
-    stack <- PEM.muxMemPred sym p (predStack predT) (predStack predF)
-    mem <- PEM.muxMemPred sym p (predMem predT) (predMem predF)
+    stack <- PEM.mux sym p (predStack predT) (predStack predF)
+    mem <- PEM.mux sym p (predMem predT) (predMem predF)
     return $ StatePred regs stack mem
 
 statePredFalse :: WI.IsExprBuilder sym => sym -> StatePred sym arch
```
travitch, line 68: This isn't necessarily part of this PR, but what is a `False` `StatePred`? One that is totally empty and covers nothing?

src/Pate/MemCell.hs
```diff
@@ -75,121 +73,87 @@ instance PC.TestEquality (WI.SymExpr sym) => Eq (MemCell sym arch w) where
 instance PC.OrdF (WI.SymExpr sym) => Ord (MemCell sym arch w) where
   compare stamp1 stamp2  = PC.toOrdering $ PC.compareF stamp1 stamp2
 
--- | This is a collection of 'MemCell' that all represent memory regions of the same size.
---
--- Each 'MemCell' is associated with the predicate that says whether or not the
+-- | Each 'MemCell' is associated with the predicate that says whether or not the
 -- described memory is contained in the 'Pate.Equivalence.MemPred'.
-newtype MemCells sym arch w = MemCells (Map.Map (MemCell sym arch w) (WI.Pred sym))
+type MemCellPred sym arch = Map.Map (Some (MemCell sym arch)) (WI.Pred sym)
```
travitch, line 78: Is this meant to be an abstraction maintained by this module? Could we make it a `newtype` to do so?

src/Pate/MemCell.hs
```diff
@@ -75,121 +73,87 @@ instance PC.TestEquality (WI.SymExpr sym) => Eq (MemCell sym arch w) where
...
 
-mapCellPreds ::
-  (WI.Pred sym -> IO (WI.Pred sym)) ->
-  MemCells sym arch w ->
-  IO (MemCells sym arch w)
-mapCellPreds f (MemCells stamps) = MemCells <$> mapM f stamps
-
-mergeMemCells ::
+dropTrivialCells ::
```
travitch, line 80: Can we define "trivial" cells?

src/Pate/MemCell.hs
```diff
@@ -75,121 +73,87 @@ instance PC.TestEquality (WI.SymExpr sym) => Eq (MemCell sym arch w) where
...
+    MapM.mergeA
+      (MapM.traverseMissing (\_ pT -> WI.andPred sym pT p))
+      (MapM.traverseMissing (\_ pF -> WI.andPred sym pF notp))
+      (MapM.zipWithAMatched (\_ p1 p2 -> WI.baseTypeIte sym p p1 p2))
+      cellsT
+      cellsF
 
 -- | True if this cell is logically equivalent to any cell in the given
 -- collection. Note that this is still false if the given cell overlaps
 -- two different entries.
```
travitch, line 131: Could we elaborate this note? Is it saying that it will return `True` if the `MemCell` overlaps just one entry in `MemCellPred`?

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -0,0 +1,224 @@
...
+      -- specified by 'memDomainLocs'.  If false, then the predicate is true
+      -- everywhere but these locations.
+      --
+      -- Currently this is always concrete, but alternate strategies
+      -- for computing pre-domains may decide to use symbolic polarities,
+      -- and there is no fundamental reason to exclude this case.
+      }
+
+-- | Map the internal 'PMC.MemCell' entries representing the locations of a 'MemoryDomain', preserving its polarity.
+traverseWithCellPar ::
```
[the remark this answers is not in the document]
danmatichuk: It depends on the 'future' parameter. We can recover normal traversal by making 'future' the same as 'm' (which is how  `traverseWithCell` is defined). I'll make a note that it is parallel when `future` is `Par.Future`.

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -0,0 +1,224 @@
...
+  OrdF (W4.SymExpr sym) =>
+  sym ->
+  [(Some (PMC.MemCell sym arch), W4.Pred sym)] ->
+  W4.Pred sym ->
+  IO (MemoryDomain sym arch)
+fromList sym l pol = do
+  -- NOTE: We can't just use Data.Map.fromList here because it will discard duplicate
+  -- entries
+  let maps = map (\(cell, p) -> M.singleton cell p) l
+  merged <- foldM (PMC.mergeMemCellPred sym) M.empty maps
```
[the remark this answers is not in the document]
danmatichuk: I don't think there is a monadic variant readily available, which is necessary since we need to construct a disjunction if there are key clashes.

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -0,0 +1,224 @@
...
+  return $ MemoryDomain merged pol
+
+-- | Similar to 'fromList' but assumes the cells are in order and disjoint
+fromAscList ::
+  W4.IsExprBuilder sym =>
+  OrdF (W4.SymExpr sym) =>
+  [(Some (PMC.MemCell sym arch), W4.Pred sym)] ->
+  W4.Pred sym ->
+  MemoryDomain sym arch
+fromAscList l pol = MemoryDomain (M.fromAscList l) pol
```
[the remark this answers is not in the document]
danmatichuk: Possibly not, it's only used internally at the moment to avoid rebuilding the map if it can be avoided.

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -0,0 +1,224 @@
...
+  W4.Pred sym ->
+  MemoryDomain sym arch ->
+  MemoryDomain sym arch ->
+  IO (MemoryDomain sym arch)
+mux sym p predT predF = case W4.asConstantPred p of
+  Just True -> return predT
+  Just False -> return predF
+  _ -> do
+    pol <- W4.baseTypeIte sym p (memDomainPolarity predT) (memDomainPolarity predF)
+    locs <- PMC.muxMemCellPred sym p (memDomainPred predT) (memDomainPred predF)
```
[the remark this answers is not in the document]
danmatichuk: Actually I realized that this invariant doesn't hold, since our "external" domains flip the polarity of the stack to demand exact equivalence. I'll write up a ticket about it and update the comment, but no we can't just assert that the polarity is always concrete.

src/Pate/Equivalence/MemoryDomain.hs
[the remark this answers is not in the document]
danmatichuk: I'll drop it from the interface.

src/Pate/Equivalence/StatePred.hs
danmatichuk: Yes it's empty, there is another branch in progress which tries to standardize some of the language (renaming `StatePred` to EquivalenceDomain and `statePredFalse` is then just `empty`)

src/Pate/MemCell.hs
danmatichuk: Yes, although it is currently violated in a couple places that would be cumbersome to refactor.  I've wrapped it in a `newtype` an put up a ticket (#207) to clean up that boundary,

src/Pate/MemCell.hs
danmatichuk: Elaborated this a bit.

travitch (approved):

PR #203, opened 2022-03-09
Abstract interpretation and strongest posts

robdockins:
This is currently a very preliminary take on the new strongest postcondition setup we have been discussing.


robdockins:
The main changes that are not confined to the new `PairGraph` module have to do with adding system call observables to the trace memory model, which necessitates some rearranging of how other code computes, e.g., footprints and the `validConcreteReads` operation.

src/Pate/Memory/MemTrace.hs
```diff
@@ -500,22 +506,43 @@ instance TestEquality (SymExpr sym) => Eq (MemOp sym ptrW) where
...
+  --   later events appear closer to the front of the sequence.
   , memState :: MemTraceState sym ptrW
   -- ^ the logical contents of memory
   }
 
 data MemTraceState sym ptrW = MemTraceState
   { memArr :: MemTraceArr sym ptrW }
 
-type MemTraceSeq sym ptrW = Seq (MemOp sym ptrW)
+type MemTraceSeq sym ptrW = SymSequence sym (MemEvent sym ptrW)
```
travitch, line 545: Ah yes - I remember now. I had an experimental branch where I made this change too (but it was in a copy of the memory model I was experimenting with). I think this is entirely reasonable.

src/Pate/Memory/MemTrace.hs
```diff
@@ -500,22 +506,43 @@ instance TestEquality (SymExpr sym) => Eq (MemOp sym ptrW) where
...
+        (compare vr1 vr2) <>
+        (toOrdering $ compareF vo1 vo2) <>
+        compare end1 end2
+
+data MemEvent sym ptrW where
+  MemOpEvent :: MemOp sym ptrW -> MemEvent sym ptrW
+  SyscallEvent :: forall sym ptrW w.
+    -- TODO, something more realistic
+    SymExpr sym (BaseBVType w) ->
+    MemEvent sym ptrW
```
travitch, line 532: What do you think about making a separate module to define observable events? I suspect we might want to generate them in places besides the memory model (especially when we start adding support for e.g., calls to external functions)

src/Pate/Memory/MemTrace.hs
```diff
@@ -703,6 +715,63 @@ execMacawStmtExtension (MacawArchEvalFn archStmtFn) mkundef mvar globs stmt
       off' <- bvZext sym w offset
       pure (LLVMPointer region off')
 
+newtype MacawSyscallModel sym arch = MacawSyscallModel ()
```
travitch, line 718: Could you add a note on what you envision this being in the long term? Might this be something like functions implementing the models for various system calls?

src/Pate/Memory/MemTrace.hs
travitch: For reference, that PR is here https://github.com/GaloisInc/pate/pull/174/files#diff-f9510819c7bea75a6da4de8da706d3d6767f09b84e4539040629aef5a4b369dc. I tried to split up this monolithic MemTrace.hs module into more logically separate components

src/Pate/Memory/MemTrace.hs
```diff
@@ -703,6 +715,63 @@ execMacawStmtExtension (MacawArchEvalFn archStmtFn) mkundef mvar globs stmt
...
+  do let simCtx = cst^.stateContext
+     let halloc = simHandleAllocator simCtx
+     let nm = "MacawSyscall"
+     fnh <- mkHandle' halloc nm atps (StructRepr rtps)
+     let FnBindings fns = simCtx ^. functionBindings
+     let ovr  = mkOverride' nm (StructRepr rtps) $
+                  do RegMap args <- getOverrideArgs
+                     applySyscallModel rtps args syscallModel mvar
+     let fns' = FnBindings (insertHandleMap fnh (UseOverride ovr) fns)
+     return (HandleFnVal fnh, cst & stateContext . functionBindings .~ fns')
```
travitch, line 738: In general, is it okay to re-install the same handler multiple times? I guess it is just basically the same handler under many different handles

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,878 @@
...
+import qualified Pate.SimState as PS
+import qualified Pate.Solver as PS
+import qualified Pate.SimulatorRegisters as PSR
+
+import qualified Pate.Verification.Domain as PVD
+import qualified Pate.Verification.Validity as PVV
+import qualified Pate.Verification.SymbolicExecution as PVSy
+
+
+newtype Gas = Gas Word32
```
travitch, line 92: What decrements the gas? Can we document an anticipated range of reasonable values?

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,878 @@
...
+
+
+runVerificationLoop ::
+  forall sym arch.
+  PA.ValidArch arch =>
+  EquivEnv sym arch ->
+  -- | A list of block pairs to test for equivalence. They must be the entry points of a functions.
```
travitch, line 99:
```suggestion
  -- | A list of block pairs to test for equivalence. They must be the entry points of functions.
```

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,878 @@
...
+        return (result, stats)
+
+
+-- | Temporary constant value for the gas parameter.
+--   Should make this configurable.
+initialGas :: Gas
+initialGas = Gas 5
+
+-- For now, the abstract domains we track are just exactly
+--  a "StatePredSpec", but we may change/add to this as we go
```
travitch, line 128:
```suggestion
--  a 'PE.DomainSpec', but we may change/add to this as we go
```

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,878 @@
...
+         liftIO $ W4.orPred sym bothReturn abortCase
+
+
+       let doPanic = panic Solver "checkTotality" ["Online solving not enabled"]
+
+       PS.withOnlineSolver solver saveInteraction sym $ \bak ->
+           do liftIO $ LCBO.withSolverProcess bak doPanic $ \sp -> do
+                W4.assume (W4.solverConn sp) precond
+                W4.assume (W4.solverConn sp) =<< W4.notPred sym isReturn
+                W4.assume (W4.solverConn sp) =<< W4.notPred sym isUnknown
```
travitch, line 319: Does this need to be in a new solver frame?

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,878 @@
...
+tryWidenings ::
+  [IO (WidenResult sym arch)] ->
+  IO (WidenResult sym arch)
+tryWidenings [] = return NoWideningRequired
+tryWidenings (x:xs) =
+  x >>= \case
+    NoWideningRequired -> tryWidenings xs
+    res -> return res
+
+widenPostcondition :: forall sym arch.
```
travitch, line 535: This is probably fine for now, but this is probably one of the things that needs the most documentation once we are happier with it

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,878 @@
...
+       precond <- liftIO $ do
+         asm <- PS.getAssumedPred sym asmFrame
+         eqInputs <- PE.getPredomain sym stackRegion bundle eqRel (PS.specBody preD)
+         W4.andPred sym asm eqInputs
+
+       -- traceBundle bundle "== widenPost: precondition =="
+       -- traceBundle bundle (show (W4.printSymExpr precond))
+
+       PS.withOnlineSolver solver saveInteraction sym $ \bak ->
+         do liftIO $ LCBO.withSolverProcess bak doPanic $ \sp -> do
```
travitch, line 560: Is this also the kind of thing that needs a new frame? Or is there a higher-level protocol maintained in this loop that means that we are already under a frame?

src/Pate/Verification/Domain.hs
```diff
@@ -348,8 +348,7 @@ universalDomain ::
 universalDomain sym =
   let
     regDomain =
-        (PER.update sym (\_ -> W4.falsePred sym)) (MM.sp_reg @(MM.ArchReg arch))
```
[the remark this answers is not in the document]
robdockins: This actually puts the stack pointer back _in_. For the dataflow algorithm, it means we will assume SP is the same until we discover it actually isn't and have to widen. For the triple verifier, it means that some cases will have slightly stricter conditions to attempt to prove; I'm still not 100% I understand exactly when those conditions arise.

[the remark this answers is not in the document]
travitch: Oh, I understand what I am seeing now. Okay

src/Pate/Verification/PairGraph.hs
robdockins: `withOnlineSolver` starts a fresh solver instance .  If we were using a more persistent solver, then yes, I think a solver frame here would be necessary.

src/Pate/Verification/PairGraph.hs
robdockins: Agreed

src/Pate/Memory/MemTrace.hs
travitch: No... the two dimensional array memory model idea did not pan out for performance reasons.

Maybe I'll resurrect that module split later

src/Pate/Verification/PairGraph.hs
robdockins: Same situation as above, basically.  This is a fresh solver, and the slice preconditions need to be in scope for everything we are doing below.

src/Pate/Memory/MemTrace.hs
robdockins: Right now, we're making a fresh name each time, so there's no conflict.  We could cache the handlers and reuse them, but I'm not sure it is worth the trouble.

src/Pate/Verification/PairGraph.hs
robdockins: Right now, gas is used when you do a widening step along a particular control-flow path.  So, it should only be consumed as you traverse loops or recursive call cycles.

src/Pate/Memory/MemTrace.hs
```diff
@@ -1287,10 +1332,132 @@ instance OrdF (SymExpr sym) => Ord (MemFootprint sym ptrW) where
...
+   f rec (SymSequenceMerge _ p xs ys) =
+     do xs' <- rec xs
+        ys' <- rec ys
+        muxSymSequence sym p xs' ys'
+
+
+-- | Get an unordered collection of all the read memory operations
+--   that occurred in this trace memory. Control-flow conditions are
+--   dropped while constructing this set, so the conditions included
+--   on these `MemOp` values may not be accurate.
```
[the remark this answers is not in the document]
robdockins: We could, but it's a little tricky because of how it interacts with the cache. Currently the only use site for this ignores the conditions, so it seems like a waste.  I would have preferred to move the entire `validConcreteReads` function inside this module, but that creates and awkward module import cycle.

I'll comment the situation more throughly.

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,887 @@
...
+                    -- executions listed in "exits"
+                    do let oRegs  = PS.simRegs (PS.simOutState (PPa.pOriginal (PS.simOut bundle)))
+                       let pRegs  = PS.simRegs (PS.simOutState (PPa.pPatched  (PS.simOut bundle)))
+                       let oIPReg = oRegs ^. MM.curIP
+                       let pIPReg = pRegs ^. MM.curIP
+                       let oBlockEnd = PS.simOutBlockEnd (PPa.pOriginal (PS.simOut bundle))
+                       let pBlockEnd = PS.simOutBlockEnd (PPa.pPatched  (PS.simOut bundle))
+
+                       -- Ugh, gross
+                       q1 <- W4.groundEval evalFn (LCS.unRV (oBlockEnd ^. _1))
```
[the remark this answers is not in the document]
robdockins: Oh, that seems much nicer.

src/Pate/Verification/PairGraph.hs
```diff
@@ -0,0 +1,887 @@
...
+  withSym $ \sym ->
+    do vcfg <- asks envConfig
+       eqRel <- asks envBaseEquiv
+       stackRegion <- asks (PMC.stackRegion . envCtx)
+
+       let solver = PCfg.cfgSolver vcfg
+       let saveInteraction = PCfg.cfgSolverInteractionFile vcfg
+
+       precond <- liftIO $ do
+         eqInputs <- PE.getPredomain sym stackRegion bundle eqRel (PS.specBody preD)
```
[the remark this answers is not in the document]
robdockins: This is an interesting point.  With this setup, I've been thinking of the variables as essentially being "bound" to their position in the graph, and the body of the spec as being "open" rather than closed as a lambda.  I think the two views can be compatible, but maybe we shouldn't mix them up...

I don't think I'll change this for the moment, but maybe we should talk about this some more.

PR #208, opened 2022-03-09
restructure EquivTriple to push SimSpec inside

### LINKED ISSUE #198: Restructure 'EquivTriple' to avoid nested 'SimSpec's
The binding structure of EquivTripleBody is inconsistent with respect to variable binding. The post-domain is closed in a SimSpec while the pre-domain is not, instead the variables of the pre-domain are bound in a SimSpec that wraps the entire EquivTripleBody. This doesn't correctly reflect the scoping of these variables, and instead each domain should individually capture its own variables.

danmatichuk:
free variables of the post-domain should be disjoint
from the pre-domain, so this more accurately represents
the variable scope

fixes #198

### DIFF
--- a/src/Pate/Proof.hs
+++ b/src/Pate/Proof.hs
@@ -108,9 +108,9 @@ import qualified What4.Interface as W4
 -- proof objects
 
 
--- | An triple representing the equivalence conditions for a block pair. Abstracted
--- over the initial machine state. This is analogous to a 'ProofExpr' of type
--- 'ProofTriple', but contains the 'PE.StatePred' type that the verifier uses.
+-- | An triple representing the equivalence conditions for a block pair.
+-- This is analogous to a 'ProofExpr' of type
+-- 'ProofTriple', but contains the 'PE.StatePredSpec' type that the verifier uses.
 -- This redundancy can potentially be eliminated at some point.
 data EquivTriple sym arch where
   EquivTriple ::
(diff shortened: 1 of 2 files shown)


src/Pate/Proof.hs
```diff
@@ -109,32 +108,23 @@ import qualified What4.Interface as W4
 -- proof objects
 
 
--- | The body of an 'EquivTriple'.
-data EquivTripleBody sym arch where
-  EquivTripleBody ::
+-- | An triple representing the equivalence conditions for a block pair. Abstracted
+-- over the initial machine state. This is analogous to a 'ProofExpr' of type
+-- 'ProofTriple', but contains the 'PE.StatePred' type that the verifier uses.
```
travitch, line 113:
```suggestion
-- 'ProofTriple', but contains the 'PE.StatePredSpec' type that the verifier uses.
```

travitch (approved):
I'm not going to claim to fully understand what is going on here, but it seems simpler than before so that is at least good... I was surprised to see the return type of `topLevelTriple` change so much syntactically, but I guess there was a lot of complexity hidden behind that former `EquivTriple` type alias that was hard to see before.

src/Pate/Proof.hs
danmatichuk: It's probably more accurate to say that `toplevelTriple` was previously strange as a result of trying to work around the odd structure of `EquivTriple`, and now it's a bit more straightforward.

PR #211, opened 2022-03-11
identify 'StatePred' and 'EquivalenceDomain'

### LINKED ISSUE #182: De-duplicate and concretize proof domains
The proof domains have two representations as a StatePred.StatePred and as a Proof.ProofDomain. The latter is used as a node in the proof tree, and specifically supports being "grounded" to a model from the solver.

This is probably more abstraction than is necessary, and results in duplication/conversion between these representations. In anticipation of refactoring the proof structure, a good step forward is to remove this duplication and instead unify these representations.

danmatichuk:
Avoids duplicated representations for equivalence domains between the verifier and the proof structure

Fixes #182


src/Pate/Proof/Instances.hs
```diff
@@ -225,17 +226,17 @@ instance forall sym arch tp. (PA.ValidArch arch, PSo.ValidSym sym) => PP.Pretty
...
+    , PP.indent 4 $ prettyMem (PED.eqDomainStackMemory prf)
+    , "Global Memory:" <+> ppPolarity (PEM.memDomainPolarity $ PED.eqDomainGlobalMemory prf)
+    , PP.indent 4 $ prettyMem (PED.eqDomainGlobalMemory prf)
     ]
     where
-      prettyRegs = PP.vsep (map ppReg (collapseRegState (Proxy @sym) (PF.eqDomainRegisters prf)))
+      prettyRegs = PP.vsep (map ppReg (collapseRegState (Proxy @sym) (PED.eqDomainRegisters prf)))
 
       ppReg :: (Some (MM.ArchReg arch), W4.Pred sym) -> PP.Doc a
       ppReg (Some reg, p) = case W4.asConstantPred p of
```
travitch, line 242: Could this function be defined next to the register domain to preserve some abstraction?

src/Pate/Register.hs
```diff
@@ -6,9 +6,6 @@
 module Pate.Register (
```
travitch, line 6: Could we clean up these warnings?

src/Pate/Verification.hs
```diff
@@ -545,8 +545,8 @@ withPair pPair f = do
 provePostcondition ::
```
travitch, line 545: What is the meaning of this `DomainSpec`? What is it meant to be used for?

I think that the generated proof term makes sense

src/Pate/Verification/Domain.hs
```diff
@@ -5,18 +5,19 @@
...
 module Pate.Verification.Domain (
-    allRegistersDomain
-  , guessEquivalenceDomain
+    guessEquivalenceDomain
   , equateInitialStates
   , universalDomain
   , universalDomainSpec
   ) where
 
+import           Control.Lens ( (.~), (&) )
```
travitch, line 20: Can we clean up the unused import warnings down here?

src/Pate/Equivalence/EquivalenceDomain.hs
```diff
@@ -0,0 +1,124 @@
...
+  EquivalenceDomain ::
+    { -- | Each register is considered to be in this domain if the given predicate is true.
+      eqDomainRegisters :: RegisterDomain sym arch
+      -- | The memory domain that is specific to stack variables.
+    , eqDomainStackMemory :: PEM.MemoryDomain sym arch
+      -- | The memory domain that is specific to non-stack (i.e. global) variables.
+    , eqDomainGlobalMemory :: PEM.MemoryDomain sym arch
+    }  -> EquivalenceDomain sym arch
+
+type RegisterDomain sym arch = MM.RegState (MM.ArchReg arch) (Const (WI.Pred sym))
```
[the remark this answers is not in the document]
travitch: Also, food for thought, is this the `Const` that seems to leak out through the rest of the code? Is there an API tweak that might keep all of that internal to this module?

src/Pate/Equivalence.hs
```diff
@@ -93,17 +92,9 @@ instance Semigroup EquivalenceStatus where
 instance Monoid EquivalenceStatus where
   mempty = Equivalent
 
-type StatePredSpec sym arch = SimSpec sym arch (PES.StatePred sym arch)
+type DomainSpec sym arch = SimSpec sym arch (PED.EquivalenceDomain sym arch)
```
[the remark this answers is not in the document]
danmatichuk: The 'SimSpec' pattern is somewhat general and exposed here, since the inner domains need to be instantiated via 'bindSpec' before they can actually be used. The only thing I can think of that would potentially justify defining this as a 'newtype' would be to formally distinguish when a 'DomainSpec' is being interpreted as a pre-domain vs. post-domain (i.e. define PreDomain and PostDomain types which are both a 'DomainSpec' under the hood).

src/Pate/Equivalence.hs
```diff
@@ -270,55 +261,53 @@ impliesPostcondPred sym (PPa.PatchPair stO stP) stPredAsmSpec stPredConclSpec =
...
+  PED.EquivalenceDomain sym arch ->
   IO (EquivRelation sym arch, W4.Pred sym)
-getPostcondition sym bundle eqRel stPred = do
-  post <- statePredPost sym (simOutO bundle) (simOutP bundle) eqRel stPred
+getPostdomain sym bundle eqRel stPred = do
+  post <- eqDomPost sym (simOutO bundle) (simOutP bundle) eqRel stPred
   return (weakenEquivRelation sym stPred eqRel, post)
 
 -- | Weaken an equivalence relation to be conditional on exactly this predicate.
 -- This is meant to be used for reporting only.
```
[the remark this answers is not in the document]
danmatichuk: It's still a predicate, I'm not sure I understand the question. This is basically flattening everything into the exact equivalence problem that we're feeding to the solver.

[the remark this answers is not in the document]
travitch: I thought that the reference to a predicate in the comment was referring to the `StatePred` that became a `EquivalenceDomain`. That may just be a misunderstanding

src/Pate/Proof/Instances.hs
danmatichuk: Do you mean 'collapseRegState' in particular, or the pretty printing logic? I've wrapped RegisterDomain in a newtype now, so converting it to a list of register-predicate pairs now at least has to go through that interface.

travitch: I was thinking about the pretty printing logic for registers in general. If there was just a pretty printer (i.e., this pretty printer) for the domain registers (defined with that newtype), this would just be `ppDomainRegisters` and you don't even need to expose the conversion to a list.

src/Pate/Equivalence.hs
```diff
@@ -201,7 +192,7 @@ stateEquivalence hdr sym stackRegion =
   in EquivRelation (registerEquivalence hdr sym) stackEq memEq
 
 -- | Resolve a domain predicate and equivalence relation into a precondition
```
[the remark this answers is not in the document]
danmatichuk: Added a bit of explanation.

src/Pate/Equivalence.hs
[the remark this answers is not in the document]
danmatichuk: Oh I understand now, yes you're right.

src/Pate/Equivalence/EquivalenceDomain.hs
```diff
@@ -0,0 +1,124 @@
...
+  Just True -> return domT
+  Just False -> return domF
+  _ -> do
+    regs <- PRt.zipWithRegStatesM (eqDomainRegisters domT) (eqDomainRegisters domF)
+      (\_ (Const pT) (Const pF) -> Const <$> WI.baseTypeIte sym p pT pF)
+    stack <- PEM.mux sym p (eqDomainStackMemory domT) (eqDomainStackMemory domF)
+    mem <- PEM.mux sym p (eqDomainGlobalMemory domT) (eqDomainGlobalMemory domF)
+    return $ EquivalenceDomain regs stack mem
+
+containsReg ::
```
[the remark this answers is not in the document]
danmatichuk: Renamed

src/Pate/Verification.hs
danmatichuk: Added some documentation.

src/Pate/Proof/Instances.hs
danmatichuk: Sure I can shuffle that into the RegisterDomain module (probably more examples in that module now that can be moved closer to their respective types). This logic still ends up somewhat duplicated in the Render.Proof module though, since it doesn't want to just produce a PP.Doc.

travitch: ah that is fair. The duplication in the UI is truly unfortunate and I wish I had a better answer there

travitch (approved):

PR #222, opened 2022-03-21
Speed up web UI by avoiding processing redundant proof events

### LINKED ISSUE #221: Web interface is unresponsive on larger binaries
Often the web interface will hang when the verifier is working on larger binaries. So far two reasons have been identified: 

1. the front-end is being redrawn for all verifier events, despite the vast majority of events not affecting the output
2. intermediate proof events are not being streamed out during the proof, and instead are all dumped out at once when the proof is  finished

danmatichuk:
* Only request UI updates when an event is raised that affects the output.
* Emit proof events once, and as soon as they are ready (rather than waiting for the proof to complete).
Fixes #221

### DIFF
--- a/src/Pate/Interactive.hs
+++ b/src/Pate/Interactive.hs
@@ -14,7 +14,7 @@ module Pate.Interactive (
 import qualified Control.Concurrent as CC
 import           Control.Lens ( (^.), (%~), (&), (.~) )
 import qualified Control.Lens as L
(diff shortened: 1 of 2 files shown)


travitch (approved):
When merging, can you just make sure that the PR description mentions that the high-level idea of the change is that it only requests that the UI update when interesting information has changed (rather than on *any* message, as before)?

PR #235, opened 2022-03-23
Add a configuration option to let users restrict the scope of verific…

### LINKED ISSUE #215: Add a mechanism to restrict the scope of the verification
For larger programs, we need the ability to let a user specify that they don't want results for part of the program, and cause the verifier to ignore those parts.

travitch:
…ation

This change adds two options to the configuration file:
- `ignore-original-functions`
- `ignore-patched-functions`

These let the user specify a list of addresses that will be "ignored" in the
original (respectively patched) binary if they are functions that are called.
Operationally, a function is ignored by treating it as a no-op (and thus
ignoring all of its callees). This is unsafe, but can be useful for exploring
the impact of changes on large binaries.

Note that the while the two lists are independent, they probably need to align
very closely. They are separate to allow for minor code motion differences
between the two binaries, but if a function is ignored in one, it almost
certainly has to be ignored in the other, too.

Fixes #215


travitch:
This builds on the TOML config branch. Also note that there was some unfortunate code motion wrapped up in this patch. It became necessary to untangle a module import cycle. The main changes are in `parsedFunctionContaining`

danmatichuk (approved):

PR #241, opened 2022-03-23
Strongest postconditions, phase 2

robdockins:


robdockins:
Current status: some things work.  The current output generated on a particular example can be seen by running:

```
cabal run pate -- -o tests/aarch32/test-write-reorder-call.original.exe -p tests/aarch32/test-write-reorder-call.patched.exe --strongestpost -V Debug --log-file pate.log
```

This will produce the following out put in `pate.log`.  In it, you can see that it found 2 interesting situations. First, there is a totality check failure which shows diverging control flow.  This turns out to be spurious, but only because of the particular initial values of  mutable memory for this program.  The second situation shows that two writes to heap memory  have been reordered, which is potentially a real situation of interest.  Along the way, it was necessary to widen the abstract domain with respect to the value of some registers.

```
Warnings during ELF loading:
Loaded original and patched binaries
0x100d8: Simulating original blocks
0x100d8: Discarding edges: []
0x100d8: Simulating patched blocks
0x100d8: Discarding edges: []
0x100d8: Finished simulating blocks
0x100d8: Discovering exit pairs from ConcreteAddress 0x100d8 vs. ConcreteAddress 0x100d8
0x100d8: 1 pairs found!
0x100d8: Observables agree
0x100d8: Found extra exit while checking totality:
0 function return Just (0x4,"0x10108: LDM_A1(xxxxxxxx.xxxxxxxx.10x1xxxx.xxxx1000) Rn 13, W 1, cond 14, register_list 34816")
100fc function call Just (0x10,"0x100e8: BL_i_A1(xxxxxxxx.xxxxxxxx.xxxxxxxx.xxxx1011) cond 14, imm24 3")

0x100d8: Handling proof case 0
0x100d8:   targetCall: ConcreteAddress 0x10114
0x100d8:   Return target ConcreteAddress 0x10104, ConcreteAddress 0x100ec
0x100d8:   Is Syscall? False
0x100d8: == Found a widening, returning into the loop ==
0x100d8: PSTATE_N PSTATE_V _R14 _R3
0 memory locations

0x100d8: == Found a widening, returning into the loop ==
0x100d8: PSTATE_C PSTATE_Z _R2
0 memory locations

0x100d8: Successfully widened postcondition
0x10114: Simulating original blocks
0x10114: Discarding edges: []
0x10114: Simulating patched blocks
0x100fc: Discarding edges: []
0x10114: Finished simulating blocks
0x10114: Discovering exit pairs from ConcreteAddress 0x10114 vs. ConcreteAddress 0x100fc
0x10114: 0 pairs found!
0x10114: Obserables disagree!
0x10114: == Original sequence ==
0x10114: Write 0x800000:[32] <- 0x1:[32] 
Write 0x0:[32] <- 0x2:[32] 
0x10114: == Patched sequence ==
0x10114: Write 0x0:[32] <- 0x2:[32] 
Write 0x800000:[32] <- 0x1:[32] 
0x10114: Totality check succeeded.
```

travitch:
Very cool!

src/Pate/Memory/MemTrace.hs
```diff
@@ -654,7 +684,9 @@ execMacawStmtExtension (MacawArchEvalFn archStmtFn) mkundef syscallModel mvar gl
     MacawArchStmtExtension archStmt -> archStmtFn mvar globs archStmt
 
     MacawArchStateUpdate{} -> \cst -> pure ((), cst)
-    MacawInstructionStart{} -> \cst -> pure ((), cst)
+    MacawInstructionStart _baddr iaddr dis ->
+      liftToCrucibleState mvar $ \sym ->
+        modify (\mem -> mem{ memCurrentInstr = toMuxTree sym (Just (iaddr,dis)) })
```
robdockins, line 689: @travitch, Not sure I'm grabbing the right data here, sometimes the results I get are a bit strange.  What's the difference between the "block address" and the "instruction address" here?

robdockins:
This is currently in an OK-ish intermediate state, and I think it's a good time to do a checkpoint merge. There are some issues with the widening procedure I haven't nailed down yet, but I think that can wait until after this merge.

The results from running this on the challenge03 variant are as shown below.  It correctly captures the new "observable" write to global memory, but fails to correctly handle a `bxeq` instruction which I think is masking additional interesting things it might find.  It also fails to converge on address 0x116dc.  I am not entirely sure, but I think this corresponds to the unrolling of a `FD_ZERO` macro, and I'm not sure why it has trouble with it; it may be related to the widening problems.

```
 cabal run pate -- -o demos/challenge03-variant/challenge03.orig.exe -p demos/challenge03-variant/challenge03.patched.exe -V Debug --log-file pate.log -b demos/challenge03-variant/challenge03.toml --strongestpost

ConcreteAddress 0x11a5c observable sequences disagree
Original sequence:
  Write 0x80000005:[32] <- 0x1:[8]
Patched sequence:
  Write 0x80000005:[32] <- 0x1:[8]
  Write 0x80000006:[32] <- 0x1:[8]
ConcreteAddress 0x11e7c program control flow desyncronized
  Original: 0x11e88 syscall Just (0x8,"0x11e84: BX_A1(0001xxxx.IIIIIIII.0010IIII.xxxx0001) Rm 14, cond 0, QuasiMask0 QuasiMask\"(\"12): 4095")
  Patched:  0x11e88 syscall Just (0x8,"0x11e84: BX_A1(0001xxxx.IIIIIIII.0010IIII.xxxx0001) Rm 14, cond 0, QuasiMask0 QuasiMask\"(\"12): 4095")
GraphNode ConcreteAddress 0x116dc vs. ConcreteAddress 0x116dc analysis failed to converge
```

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -0,0 +1,917 @@
...
+         [ pretty pPair <+> pretty "observable sequences disagree"
+         , pretty "Original sequence:"
+         ] ++
+         [ indent 2 (MT.prettyMemEvent ev) | ev <- oEvs ] ++
+         [ pretty "Patched sequence:" ] ++
+         [ indent 2 (MT.prettyMemEvent ev) | ev <- pEvs ]
+
+   reportDesync (pPair, TotalityCounterexample (oIP, oEnd, oInstr) (pIP, pEnd, pInstr)) =
+     liftIO $ putStrLn $ show $ vcat $
+       [ pretty pPair <+> pretty "program control flow desyncronized"
```
travitch, line 162:
```suggestion
       [ pretty pPair <+> pretty "program control flow desynchronized"
```

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -0,0 +1,917 @@
...
+    Nothing -> return gr
+    Just (gr', nd, d) ->
+      do gr'' <- visitNode nd d gr'
+         pairGraphComputeFixpoint gr''
+
+-- | Perform the work of propagating abstract domain information through
+--   a single program "slice". First we check for equivalence of observables,
+--   then we check the totality of the computed exits, then we update any
+--   function return nodes, and finally we follow all the exit nodes
+--   that were computed via "discoverPairs."
```
travitch, line 190: Is this ordering important?

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -0,0 +1,917 @@
...
+    Just _ -> return gr
+    Nothing ->
+      do res <- doCheckObservables asm bundle preD
+         case res of
+           ObservableCheckEq ->
+             do traceBundle bundle "Observables agree"
+                return gr
+           ObservableCheckError msg ->
+             do traceBundle bundle ("Error checking observables: " ++ msg)
+                return gr
```
travitch, line 298: Should this error be recorded somewhere? Or is that unimportant for some reason?

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -0,0 +1,917 @@
...
+         liftIO (W4.orPred sym isJump =<< W4.orPred sym isFail isBranch)
+
+       -- TODO, I really don't understand this abort case stuff, but it was copied
+       -- from the triple verifier.
+       isReturn <- do
+         bothReturn <- PD.matchingExits bundle MS.MacawBlockEndReturn
+         abortO <- PAb.isAbortedStatePred (PPa.getPair @PBi.Original (simOut bundle))
+         returnP <- liftIO $ MS.isBlockEndCase (Proxy @arch) sym (PS.simOutBlockEnd $ PS.simOutP bundle) MS.MacawBlockEndReturn
+         abortCase <- liftIO $ W4.andPred sym abortO returnP
+         liftIO $ W4.orPred sym bothReturn abortCase
```
travitch, line 556: @danmatichuk This would be a helpful thing for you to comment on

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -0,0 +1,917 @@
...
+  EquivM sym arch (Maybe BS.ByteString)
+findPLTSymbol blkO blkP =
+  do PA.SomeValidArch archData <- asks envValidArch
+     let oExtraMap = PA.validArchOrigExtraSymbols archData
+     let pExtraMap = PA.validArchPatchedExtraSymbols archData
+
+     let oAddr = PAd.addrToMemAddr (PB.concreteAddress blkO)
+     let pAddr = PAd.addrToMemAddr (PB.concreteAddress blkP)
+
+     case (MM.asAbsoluteAddr oAddr, MM.asAbsoluteAddr pAddr) of
```
travitch, line 782: I believe this will always fail for position-independent executables. Depending on what exactly you need here, this might be better as `MM.segmentOffAddr`

src/Pate/Memory/MemTrace.hs
robdockins: Ah, that makes a lot more sense.

src/Pate/Verification.hs
```diff
@@ -169,6 +169,7 @@ verifyPairs ::
 verifyPairs validArch logAction elf elf' vcfg pd = do
   Some gen <- liftIO N.newIONonceGenerator
   sym <- liftIO $ WE.newExprBuilder WE.FloatRealRepr WE.EmptyExprBuilderState gen 
+  liftIO $ WE.startCaching sym
```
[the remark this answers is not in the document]
robdockins: We can back this out maybe, I'm not sure it's really "necessary." However, it seems like it might be helpful in our setting, and doesn't seem to hurt things.

Basically, my assumption is that because we are considering pretty small chunks of program at at time, the terms shouldn't get too big and the benefit of hash-consing is probably worth the overhead, but I haven't really tested that hypothesis much.

src/Pate/Verification/PairGraph.hs
```diff
@@ -5,89 +5,55 @@
...
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 module Pate.Verification.PairGraph
   ( Gas(..)
+  , initialGas
   , PairGraph(..)
```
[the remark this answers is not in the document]
robdockins: That is indeed my eventual goal. It might be possible now, but I haven't tried yet.

src/Pate/Verification/PairGraph.hs
```diff
@@ -100,60 +66,166 @@ import qualified Pate.Verification.SymbolicExecution as PVSy
...
+  , pairGraphWorklist :: !(Set (GraphNode arch))
+
+    -- | The set of blocks where this function may return to. Whenever we see a function
+    --   call to the given FunPair, we record the program point pair where the function
+    --   returns to here. This is used to tell us where we need to propagate abstract domain
+    --   information when visiting a ReturnNode.
+  , pairGraphReturnVectors :: !(Map (PPa.FunPair arch) (Set (PPa.BlockPair arch)))
+
+    -- TODO, I'm not entirely sure I love this idea of tracking error conditions in this
+    -- data structure like this.  It works for now, but maybe is worth thinking about some more.
```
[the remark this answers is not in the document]
robdockins: Indeed, streaming the results elsewhere may be a better design in the end, I'm not sure.  Because of the monotonicity of the entire system, we should never have to retract results, so they can be reported as soon as they are found.

src/Pate/Verification/PairGraph.hs
```diff
@@ -100,60 +66,166 @@ import qualified Pate.Verification.SymbolicExecution as PVSy
...
+
+    -- | If we find a counterexample regarding observables in a particular block, record it here.
+    --   Later, this can be used to generate reports to the user.  We also avoid checking for
+    --   additional counterexamples for the given block if we have already found one.
+  , pairGraphObservableReports :: !(Map (PPa.BlockPair arch) (ObservableCounterexample sym (MM.ArchAddrWidth arch)))
+
+    -- | If we find a counterexample to the exit totality check, record it here.  This occurs when
+    --   the programs have sufficently-different control flow that they cannot be syncrhonized, or
+    --   when the analysis encounters some control-flow construct it doesn't know how to handle.
+    --   Once we find a desyncrhonization error, we do not look for additional ones.
```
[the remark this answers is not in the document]
robdockins: It just means we don't look for/record additional desynchronization errors.  Abstract domains will still be propagated.

src/Pate/Verification/StrongestPosts.hs
robdockins: I don't think so.

src/Pate/Verification/StrongestPosts.hs
robdockins: I should be, yes. I think the big question is deciding how severe this class of error is. At first blush, I think these should be recoverable errors (i.e., continue the analysis) but they should definintely be recorded somewhere.  Similar with errors due to widening, etc.

src/Pate/Verification/StrongestPosts.hs
robdockins: I admit I don't fully understand how the data in the "extra symbols" map corresponds to what I have available here; if there's a better way to do this, I'd love to know.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+--
+--   This is done by looking up the abstract domain for the target node
+--   and proving that the poststate of symbolic execution satisfies
+--   the abstract domain of the target node, assuming the abstract domain of
+--   the source node.  If this is not the case, we use the resulting counterexample
+--   to determing what additional locations need to be added to the target
+--   abstract domain. We perform these widening steps in a loop until
+--   the process converges or we run out of gas.
+--
+--   When widening, we first consider register values to widen, then we look at
```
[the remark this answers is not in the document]
robdockins: I don't think the order matters in principle, but the iterations in the widening loop are bounded, so I guess it might matter in practice.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+--   consider first locations that differ in the prestate, then locations that
+--   were written during the execution of the block.  In theory, this should
+--   enough to account for all the sources of differences we have to consider.
+--
+--   If, for some reason, we cannot find appropraite locations to widen, we
+--   widen as much as we can, and report an error.
+widenAlongEdge ::
+  SimBundle sym arch {- ^ results of symbolic execution for this block -} ->
+  GraphNode arch {- ^ current graph node -} ->
+  AbstractDomain sym arch {- ^ current abstract domain -} ->
```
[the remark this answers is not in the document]
robdockins: Yes, this is the source abstract domain.  The target domain is looked up, as you say.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+
+instance (OrdF (W4.SymExpr sym), PA.ValidArch arch) => Semigroup (WidenLocs sym arch) where
+  (WidenLocs r1 m1) <> (WidenLocs r2 m2) = WidenLocs (r1 <> r2) (m1 <> m2)
+
+instance (OrdF (W4.SymExpr sym), PA.ValidArch arch) => Monoid (WidenLocs sym arch) where
+  mempty = WidenLocs mempty mempty
+
+data WidenResult sym arch
+  = NoWideningRequired
+  | WideningError String (WidenLocs sym arch) (AbstractDomain sym arch)
```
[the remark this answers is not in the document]
robdockins: Running out of gas is one.  A more serious one is when the solver tells us that the post-domain isn't wide enough, but we can't figure out what to do about it.

src/Pate/Verification.hs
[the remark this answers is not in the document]
travitch: I'm okay keeping it. I'd just like a comment to that effect (i.e., that correctness does not depend on it)

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+            widenLoop sym bak initialGas eqCtx postD0 Nothing
+
+ where
+   doPanic = panic Solver "widenPostcondition" ["Online solving not enabled"]
+
+   -- TODO, we should probably have some way to bound the amout of times we can
+   --  recurse into the widening loop, or we really need to be very careful to
+   --  make sure that this kind of local widening will terminate in a reasonable
+   --  number of steps.
+   widenLoop ::
```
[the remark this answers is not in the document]
robdockins: That's correct.

I feel like the strategies listed here ought to be enough to always find a widening, but it isn't working in some cases and I'm still not sure why yet.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+     , sym ~ W4.ExprBuilder t st fs
+     , W4.OnlineSolver solver
+     , LCB.IsSymBackend sym bak
+     , PA.ValidArch arch ) =>
+     sym ->
+     bak ->
+     Gas ->
+     EquivContext sym arch ->
+     AbstractDomain sym arch ->
+     Maybe (WidenLocs sym arch) ->
```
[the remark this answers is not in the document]
robdockins: These are locations where we have already widened on previous iterations.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+                 Sat evalFn ->
+                   -- The current execution does not satisfy the postcondition, and we have
+                   -- a counterexample.
+                   widenUsingCounterexample sym evalFn bundle eqCtx postCondAsm postCondStatePred preD mlocs postD
+
+        -- Re-enter the widening loop if we had to widen at this step.
+        -- If this step completed, use the success continuation to
+        -- return the result.  Only in the first iteration will we
+        -- finally return a `NoWideningRequired`.  In other cases, we
+        -- had to widen at least once.
```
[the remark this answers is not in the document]
robdockins: Ah, this comment refers to an earlier version of this code.  I'll update the comment.

Basically, on the first iteration, we have a `Nothing` as the argument that tells us what widening has already been done; if no additional widening is required, then no widening was require at all, and we don't need to update the abstract domain for this program point.  If no widening was in _this iteration_, but we have already done widening in previous iterations, then the abstract domain does require widening.  If widening was required this iteration, then we reenter the loop.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+               traceBundle bundle (show (PEE.ppEquivalenceDomain W4.printSymExpr (PS.specBody preD)))
+               traceBundle bundle "===== POSTDOMAIN ====="
+               traceBundle bundle (show (PEE.ppEquivalenceDomain W4.printSymExpr (PS.specBody postD')))
+-}
+               return res
+
+          NoWideningRequired ->
+            case mlocs of
+              Nothing   -> return NoWideningRequired
+              Just locs -> return (Widen locs postD)
```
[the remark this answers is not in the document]
robdockins: This is the case that terminates the loop. In the `Just` case, we had to make some widening change in a previous iteration.

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,529 @@
...
+
+          Widen locs postD' ->
+            do traceBundle bundle "== Found a widening, returning into the loop =="
+               traceBundle bundle (show locs)
+               let newlocs = case mlocs of
+                               Nothing    -> Just locs
+                               Just locs' -> Just (locs <> locs')
+               widenLoop sym bak (Gas (i-1)) eqCtx postD' newlocs
```
robdockins, line 307: Notice in this case `newlocs` is always a `Just`

src/Pate/Verification.hs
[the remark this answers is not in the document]
danmatichuk: NB: the current verifier uses caching as well

src/Pate/Verification/Widening.hs
```diff
@@ -0,0 +1,571 @@
...
+  sym ->
+  W4.GroundEvalFn t ->
+  SimBundle sym arch ->
+  EquivContext sym arch ->
+  W4.Pred sym ->
+  PEE.EquivalenceDomain sym arch ->
+  AbstractDomain sym arch ->
+  AbstractDomain sym arch ->
+  IO (WidenResult sym arch)
+-- TODO? should we be using postCondAsm and postConstStatePred?
```
danmatichuk, line 407: I think you want to use `postCondStatePred` to avoid considering locations as "unequal" if they aren't included in the target post-domain.

danmatichuk: In fact you may want to pass in the intermediate `MemoryCondition` instead, which captures the result of `resolveCellEquiv` or `resolveRegisterEquiv`, but conditional on being in the target domain..

src/Pate/Memory/MemTrace.hs
```diff
@@ -1575,13 +1630,31 @@ instance PEM.ExprMappable sym (MemOp sym w) where
 instance PEM.ExprMappable sym (MemEvent sym w) where
   mapExpr sym f = \case
     MemOpEvent op -> MemOpEvent <$> PEM.mapExpr sym f op
-    SyscallEvent arg -> SyscallEvent <$> f arg
+    SyscallEvent i arg -> SyscallEvent i <$> f arg -- TODO? rewrite the mux tree?
```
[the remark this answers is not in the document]
robdockins: I'm in favor of eventually removing the ExprMappable instances, but for now they are providing an important service (making sure that concrete read-only reads are serviced).

src/Pate/Memory/MemTrace.hs
```diff
@@ -1575,13 +1630,31 @@ instance PEM.ExprMappable sym (MemOp sym w) where
...
+      do x'  <- PEM.mapExpr sym f x
+         xs' <- rec xs
+         consSymSequence sym x' xs'
+    SymSequenceAppend _ xs ys ->
+     do xs' <- rec xs
+        ys' <- rec ys
+        appendSymSequence sym xs' ys'
+    SymSequenceMerge _ p xs ys ->
+     do p' <- f p
+        iteM muxSymSequence sym p' (rec xs) (rec ys)
```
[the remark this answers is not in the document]
robdockins: This turned out to be important to avoid spurious counterexamples involving observables. I think we can avoid needing this (and the whole ExprMappable rewriting infrastructure) if we are a bit more carful about how we set up the initial states.

src/Pate/Verification/Widening.hs
robdockins: Does that really make sense? This is where we are deciding what things we need to _add_ to the target domain.

robdockins:
There are certainly still questions and issues to be resolved about this work, but I think we are in a good place to merge this checkpoint.

src/Pate/Verification/Widening.hs
danmatichuk: Aren't you technically removing them from the target domain? (i.e. we are weakening the post-condition to exclude locations which aren't equal). In which case I think what you'll end up with is each widening step including locations to widen that are already excluded by the domain (i.e. have already been excluded in a previous widening iteration).

robdockins: I guess it depends on how you think about it.  In this algorithm, the domains always start equating everything, and you progressively add exceptions where things don't have to be equal, so I think about the process of widening as adding new locations.

 At any rate, I think we're already doing what needs to be done via the `assume` statement here:

https://github.com/GaloisInc/pate/blob/f91ac87946fdd10b984bd48a5a93b2985641f3d3/src/Pate/Verification/Widening.hs#L288

But maybe I'm missing something.  I don't really understand the difference between `eqDomPost` and the other things you mentioned.

PR #249, opened 2022-03-24
add intermediate representation for equality

### LINKED ISSUE #238: Retain structure when generating equivalence predicates
There are three places where structured information is flattened into a single predicate for the purposes of using the solver:

1. The pre-condition for assuming initial equivalence for a slice
2. The post-condition for establishing the final equivalence for a slice
3.  The conditional equivalence predicate that is computed as a result of an inequivalence result

In each of these cases, the flattened predicate is simply a conjunction of equality tests or path conditions for each location relevant to the slice (i.e. register or memory cell). It would be useful to retain this structure for longer so that the predicate can be more easily interpreted, rather than trying to decode the meaning of each individual conjunct.

danmatichuk:
this adds an additional step between converting a
domain into an equivalence problem, where a
'StateCondition' roughly captures the collection of
equivalence assertions that eventually are flattened
into a predicate

fixes #238

### DIFF
--- a/src/Pate/Equivalence.hs
+++ b/src/Pate/Equivalence.hs
@@ -44,6 +44,13 @@ module Pate.Equivalence
   , resolveCellEquivStack
   , registerValuesEqual
   , EquivContext(..)
+  , MemoryCondition(..)
+  , RegisterCondition(..)
+  , StateCondition(..)
+  , preCondPredicate
+  , postCondPredicate
+  , memPreCondToPred
+  , memPostCondToPred
   ) where
 
 import           Control.Lens hiding ( op, pre )
@@ -74,6 +81,8 @@ import           What4.ExprHelpers
 import qualified Pate.Equivalence.MemoryDomain as PEM
 import qualified Pate.Equivalence.RegisterDomain as PER
 import qualified Pate.Equivalence.EquivalenceDomain as PED
+
+import qualified Pate.Panic as PP
 
 data EquivalenceStatus =
     Equivalent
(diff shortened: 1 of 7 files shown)


src/Pate/Equivalence.hs
```diff
@@ -408,29 +440,64 @@ resolveCellEquivStack ::
...
+  Just False -> fF
+  _ -> do
+    MemoryCondition condT memEqT <- fT
+    MemoryCondition condF memEqF <- fF
+    if memEqT == memEqF then
+      MemoryCondition <$> PEM.mux sym p condT condF <*> pure memEqT
+    else PP.panic PP.Verifier "muxMemCond" ["Incompatible memory region conditions"]
+
+
+-- | Flatten a structured 'MemoryCondition' representing a memory post-condition into
```
travitch, line 476:
```suggestion
-- | Flatten a structured 'MemoryCondition' representing a memory pre-condition into
```

src/Pate/Equivalence.hs
```diff
@@ -408,29 +440,64 @@ resolveCellEquivStack ::
...
--- on the given domain for the given input state
--- This predicate is meant to be *assumed* true.
-memDomPre ::
+
+-- | Structurally equivalent to a 'PEM.MemoryDomain', however the predicates
+-- relate to equality on the given cells.
+-- Note in the case of a negative polarity precondition, this isn't quite true,
+-- since there is not a pointwise assertion for equality. Instead the predicate
+-- is simply the condition under which that cell should be excluded from the
+-- assumed initial equality.
```
travitch, line 449: Can you note where that logic lives? Is that `memPreCondToPred`?

src/Pate/Equivalence.hs
```diff
@@ -468,6 +564,9 @@ getRegionEquality sym memEq memO memP = case memEq of
   MemEqAtRegion stackRegion -> MT.memEqAtRegion sym stackRegion memO memP
   MemEqOutsideRegion stackRegion -> MT.memEqOutsideRegion sym stackRegion memO memP
 
+newtype RegisterCondition sym arch =
```
travitch, line 567: Can you document a bit about what this is meant to be used for?

What do the predicates in the `RegState` denote?

src/Pate/Equivalence.hs
```diff
@@ -468,6 +564,9 @@ getRegionEquality sym memEq memO memP = case memEq of
...
+  RegisterCondition { regCondPreds :: MM.RegState (MM.ArchReg arch) (Const (W4.Pred sym)) }
+
 -- | Compute a predicate that is true iff the two given states are equal with respect to
```
travitch, line 570: It seems to be more structured than a `Pred` now. Is it the case that the `Pred` for each register in the `RegisterCondition` satisfies this statement now?

travitch (approved):

PR #250, opened 2022-03-25
Implement JSON reports for differential summaries

### LINKED ISSUE #177: Extend JSON reports to capture differential summaries
Also add a more extensive test case that captures the differential summary from challenge 03

travitch:
With this change, we have a concise way to report to users both differential
summaries and inlined memory differences without requiring users to trawl
through the text output.

Note that we could probably stand to improve the formatting of formulas (it is
currently just the printed output from what4).

Fixes #177

### DIFF
--- a/tools/pate/JSONReport.hs
+++ b/tools/pate/JSONReport.hs
@@ -9,17 +9,21 @@ module JSONReport (
   , waitForConsumer
   ) where
 
-import           Control.Lens ( (^.) )
 import qualified Control.Concurrent.Async as CCA
 import qualified Control.Concurrent.Chan as CCC
+import           Control.Lens ( (^.) )
 import qualified Data.Aeson as JSON
 import qualified Data.BitVector.Sized as BVS
 import qualified Data.Foldable as F
 import qualified Data.IntervalMap.Interval as DII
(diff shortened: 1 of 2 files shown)


tools/pate/JSONReport.hs
```diff
@@ -83,9 +114,19 @@ toJSON (SomeProofEvent blks proofSym) = do
                                  , inlineCallError = err
                                  }
       return (JSON.toJSON node)
+    PF.ProofStatus (PF.VerificationFail _counterexample condEqRes)
+      | Nothing <- WI.asConstantPred (PF.condEqPred condEqRes) -> do
```
travitch, line 118: @danmatichuk I'm mostly interested in whether or not this is the right way to detect a differential summary. It seems to work. I was initially checking the `condEqAbortValid` value, but that did not seem to mean what I thought it meant.

travitch: Actually, now that I think about it, should I actually be matching on a `ProofTriple` node and checking its status (in the way that this is checking status)? That would let me capture the address of the blocks that exhibit the difference

travitch: Updated to capture the original and patched block addresses

PR #265, opened 2022-04-01
don't print duplicate cells in counter-example

danmatichuk:
we need to explicitly check for equality with
respect to the grounding environment, as the grounded
variant is what we are displaying

### DIFF
--- a/src/Pate/Proof/Instances.hs
+++ b/src/Pate/Proof/Instances.hs
@@ -51,6 +51,7 @@ module Pate.Proof.Instances
   where
 
 import           Control.Lens hiding ( op, pre )
+import           Data.List ( nubBy )
 import qualified Data.IntervalMap.Interval as DII
 import qualified Data.Kind as DK
 import           Data.Maybe ( mapMaybe, catMaybes )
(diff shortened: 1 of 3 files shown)


src/Pate/Proof/Instances.hs
```diff
@@ -448,9 +449,22 @@ ppMemCellMap ::
   MapF.MapF (PMC.MemCell grnd arch) (PF.BlockSliceMemOp grnd) ->
   PP.Doc a
 ppMemCellMap dom cells = let
-  ppCells = mapMaybe (\(Pair cell v) -> ppCellVal dom cell v) $ MapF.toList cells
+  eqEntries (Pair cell1 _) (Pair cell2 _) = eqGroundCells cell1 cell2
+  ppCells = mapMaybe (\(Pair cell v) -> ppCellVal dom cell v) $ nubBy eqEntries $ MapF.toList cells
```
travitch, line 453: How many cells do you expect here in general? How about the worst case? I'm asking because `nubBy` is infamously quadratic

danmatichuk: It's the cells that show up in the counter-example for a single slice, so maybe a dozen or so at most?

travitch (approved):

Issue #268, opened 2022-04-08
Move CFG slicing code from Macaw-symbolic to Pate


danmatichuk:
The code that performs the CFG slicing (to produce a constrained CFG representing a single block slice) is fairly pate-specific and already contains pate-specific functionality to classify block exits. As we need to add additional functionality to this code to support finer-grained classification of arch terminal statements, it makes sense to move this code into pate.

PR #271, opened 2022-04-11
Move Macaw CFG slicing to Pate and add arch terminal statement classification

### LINKED ISSUE #267: Classify information about the arch terminal statements in the macaw block exit variable
Currently macaw assigns a value to a special global indicating how the block was "exited", in order to preserve this information when the CFG is sliced up. We need to include additional information about arch-specific terminal statements in order to handle them in the verified (i.e. conditional returns on ARM).

danmatichuk:
Fixes #267 and #268


src/Pate/Proof/Instances.hs
```diff
@@ -641,11 +642,11 @@ cellInGroundDomain dom cell = case isStackCell cell of
...
-  MS.MacawBlockEndCall -> "function call"
-  MS.MacawBlockEndReturn -> "function return"
-  MS.MacawBlockEndBranch -> "branch"
-  MS.MacawBlockEndArch -> "syscall"
-  MS.MacawBlockEndFail -> "analysis failure"
+  MCS.MacawBlockEndJump -> "arbitrary jump"
+  MCS.MacawBlockEndCall -> "function call"
+  MCS.MacawBlockEndReturn -> "function return"
+  MCS.MacawBlockEndBranch -> "branch"
+  MCS.MacawBlockEndArch -> "arch-specific"
```
travitch, line 651: Since we have a typeclass here, can we now make a more informative prettyprinter here?

src/Pate/Verification/SymbolicExecution.hs
```diff
@@ -63,17 +64,14 @@ archStructRepr = do
...
-  MD.ParsedTranslateError{} -> True
-  MD.ClassifyFailure{} -> True
+isTerminalBlock :: MCS.HasArchEndCase arch => MD.ParsedBlock arch ids -> Bool
+isTerminalBlock pb =
+  let MCS.MacawBlockEnd c _ = MCS.termStmtToBlockEnd (MD.pblockTermStmt pb)
+  in case c of
+    MCS.MacawBlockEndCall -> True
+    MCS.MacawBlockEndArch -> True
+    MCS.MacawBlockEndFail -> True
+    _ -> False
```
travitch, line 74: Can you make this into a complete match (i.e., matching on all of the constructors)? It is really useful if the compiler can yell at us if we add a new one and make us think about what the behavior should be

src/Data/Macaw/CFGSlice.hs
```diff
@@ -0,0 +1,177 @@
...
+    mret' <- case mret of
+      Just addr -> do
+        ptr <- MSB.valueToCrucible $ M.RelocatableValue (M.addrWidthRepr (Proxy @(M.ArchAddrWidth arch))) (M.segoffAddr addr)
+        MSB.appAtom $ C.JustValue ptrRepr ptr
+      Nothing -> do
+        MSB.appAtom $ C.NothingValue ptrRepr
+    let repr = Ctx.empty Ctx.:> C.BVRepr knownNat Ctx.:> C.MaybeRepr ptrRepr
+    MSB.appAtom $ C.MkStruct repr (Ctx.empty Ctx.:> blendK' Ctx.:> mret')
+
+assignBlockEnd :: HasArchEndCase arch =>
```
travitch, line 113: Can you just add some notes about what `assignBlockEnd` and `blockEndAtom` are working together to accomplish?

travitch (approved):

PR #272, opened 2022-04-13
Strongest post verifier, phase 3

robdockins:


src/Pate/Verification/PairGraph.hs
```diff
@@ -226,6 +263,98 @@ getReturnVectors ::
...
   fromMaybe mempty (Map.lookup fPair (pairGraphReturnVectors gr))
 
+-- | Look up the current abstract domain for the given graph node.
+getCurrentDomain ::
+  PairGraph sym arch ->
+  GraphNode arch ->
+  Maybe (AbstractDomain sym arch)
+getCurrentDomain pg nd = Map.lookup nd (pairGraphDomains pg)
+
+-- | If an observable counterexample has not already be found
```
travitch, line 273:
```suggestion
-- | If an observable counterexample has not already been found
```

src/Pate/Verification/PairGraph.hs
```diff
@@ -226,6 +263,98 @@ getReturnVectors ::
...
+  m (PairGraph sym arch)
+considerDesyncEvent gr bPair action =
+  case Map.lookup bPair (pairGraphDesyncReports gr) of
+    -- we have already found observable event differences at this location, so skip the check
+    Just _ -> return gr
+    Nothing ->
+      do (mcex, gr') <- action
+         case mcex of
+           Nothing  -> return gr'
+           Just cex -> return gr'{ pairGraphDesyncReports = Map.insert bPair cex (pairGraphDesyncReports gr) }
```
travitch, line 307: Are these two basically just the same, but with different accessors? Should we factor out this structure with e.g., lenses to reduce duplication? Probably not a huge issue, since they are pretty simple

src/Pate/Verification/Widening.hs
```diff
@@ -231,7 +232,7 @@ widenPostcondition bundle preD postD0 =
                  -- https://github.com/GaloisInc/what4/issues/196
                  W4.push sp
                  W4.assume (W4.solverConn sp) precond
-            widenLoop sym bak initialGas eqCtx postD0 Nothing
+            widenLoop sym bak (Gas 100) eqCtx postD0 Nothing
```
travitch, line 235: Can we just lift this to a named constant at the top-level? I presume we'll want to make it a parameter eventually, but making it clearer that this is a distinct value might be useful

src/Pate/Verification/Widening.hs
```diff
@@ -374,24 +388,51 @@ widenHeap ::
...
+  [Some (PMc.MemCell sym arch)] ->
+  EquivM sym arch [Some (PMc.MemCell sym arch)]
+filterCells sym evalFn memDom = \xs -> loop xs
+  where
+    filterCell (Some c) =
+      W4.groundEval evalFn =<< PEM.containsCell sym memDom c
+
+    loop :: [Some (PMc.MemCell sym arch)] -> EquivM_ sym arch [Some (PMc.MemCell sym arch)]
+
+    loop [] = return[]
```
travitch, line 429: Is this just `filterM`?

src/Pate/Verification/Widening.hs
```diff
@@ -350,9 +357,14 @@ widenUsingCounterexample ::
...
-    [ liftIO (widenRegisters sym evalFn bundle eqCtx postCondAsm postCondStatePred postD)
-    , liftIO (widenStack sym evalFn bundle eqCtx postCondAsm postCondStatePred preD postD)
-    , liftIO (widenHeap sym evalFn bundle eqCtx postCondAsm postCondStatePred preD postD)
+    [ widenRegisters sym evalFn bundle eqCtx postCondAsm postCondStatePred postD
+
+    , widenStack sym evalFn bundle eqCtx postCondAsm postCondStatePred preD postD LocalChunkWrite
+    , widenHeap sym evalFn bundle eqCtx postCondAsm postCondStatePred preD postD LocalChunkWrite
+
+    , widenStack sym evalFn bundle eqCtx postCondAsm postCondStatePred preD postD PreDomainCell
+    , widenHeap sym evalFn bundle eqCtx postCondAsm postCondStatePred preD postD PreDomainCell
```
travitch, line 366: Is it worth a note explaining why these two cases are separate?

src/Pate/Memory/MemTrace.hs
```diff
@@ -1181,6 +1233,59 @@ readMemState sym mem ptr repr = go 0 repr
   go n (PackedVecMemRepr countRepr recRepr) = V.generateM (fromInteger (intValue countRepr)) $ \i ->
       go (n + memReprByteSize recRepr * fromIntegral i) recRepr
 
+asConcreteReadOnly :: forall sym w ptrW.
```
travitch, line 1236: Could you add a comment on any limitations this currently has?

travitch (approved):

src/Pate/Verification/PairGraph.hs
robdockins: They are basically the same. I'm not convinced it's worth abstracting using lenses for just these two... but we certainly could.

src/Pate/Verification/Widening.hs
robdockins: Oh, probably, yeah.

PR #275, opened 2022-04-20
Add constant propagation for strongest postcondition verifier

danmatichuk:


src/Pate/Verification/AbstractDomain.hs
```diff
@@ -0,0 +1,506 @@
...
+  sym ->
+  AbstractDomainBody sym arch {- ^ existing abstract domain that this is augmenting -} ->
+  (forall tp. W4.SymExpr sym tp -> m (AbsRange tp)) ->
+  PS.SimBundle sym arch ->
+  m (AbstractDomainBody sym arch, Maybe (WidenLocs sym arch))
+widenAbsDomainVals sym prev f bundle = do
+  let (PPa.PatchPair absValsO absValsP) = absDomVals prev
+  (absValsO', locsO) <- widenAbsDomainVals' sym absValsO f (PS.simOutO bundle)
+  (absValsP', locsP) <- widenAbsDomainVals' sym absValsP f (PS.simOutP bundle)
+  return $ (prev { absDomVals = PPa.PatchPair absValsO' absValsP' }, Just $ locsO <> locsP)
```
travitch, line 348: I think it would be worth describing the strategy here for readers who are not familiar with constant propagation.

src/Pate/Verification/AbstractDomain.hs
```diff
@@ -0,0 +1,506 @@
...
+  PS.SimState sym arch bin ->
+  AbstractDomainVals sym arch bin ->
+  IO (PS.AssumptionFrame sym)
+absDomainValsToAsm sym st vals = do
+  memFrame <- MapF.foldrMWithKey getCell mempty (absMemVals vals)
+  regFrame <- fmap PRt.collapse $ PRt.zipWithRegStatesM (PS.simRegs st) (absRegVals vals) $ \_ val absVal ->
+    Const <$> absDomainValToAsm sym val absVal
+  return $ memFrame <> regFrame
+  where
+    getCell ::
```
travitch, line 408: It might be nice to call this something like `accumulateCell` given that it is building up the result in a fold

src/Pate/Verification/AbstractDomain.hs
```diff
@@ -0,0 +1,506 @@
...
+  MC.RegisterInfo (MC.ArchReg arch) =>
+  sym ->
+  PS.SimState sym arch bin ->
+  AbstractDomainVals sym arch bin ->
+  IO (W4.Pred sym)
+absDomainValsToPred sym st vals = do
+  asm <- absDomainValsToAsm sym st vals
+  PS.getAssumedPred sym asm
+
+absDomainToPrecond ::
```
travitch, line 434: Presumably this is meant to be assumed somewhere?

src/Pate/Verification/AbstractDomain.hs
```diff
@@ -0,0 +1,506 @@
...
+  eqInputs <- PE.getPredomain sym bundle eqCtx (absDomEq d)
+  eqInputsPred <- PE.preCondPredicate sym (PS.simInO bundle) (PS.simInP bundle) eqInputs
+  valsPred <- do
+    let PPa.PatchPair valsO valsP = absDomVals d
+    predO <- absDomainValsToPred sym (PS.simInState $ PS.simInO bundle) valsO
+    predP <- absDomainValsToPred sym (PS.simInState $ PS.simInP bundle) valsP
+    W4.andPred sym predO predP
+  W4.andPred sym eqInputsPred valsPred
+
+absDomainToPostCond ::
```
travitch, line 452: And this is meant to be proved?

src/Pate/Verification/AbstractDomain.hs
```diff
@@ -0,0 +1,506 @@
...
+instance PEM.ExprMappable sym (AbstractDomainBody sym arch) where
+  mapExpr sym f d = do
+    domEq <- PEM.mapExpr sym f (absDomEq d)
+    vals <- PEM.mapExpr sym f (absDomVals d)
+    return $ AbstractDomainBody domEq vals
+
+ppAbstractDomainVals ::
+  AbstractDomainVals sym arch bin ->
+  PP.Doc a
+ppAbstractDomainVals _d = "<TODO>"
```
travitch, line 487: Lets either fill this in or open a ticket for it

src/Pate/Verification/Widening.hs
```diff
@@ -212,42 +214,63 @@ localWideningGas :: Gas
 localWideningGas = Gas 100
 
 
-widenPostcondition :: forall sym arch.
-  PA.ValidArch arch =>
+--TODO: this is a bit of a cludge - we should standardize how EquivM interacts
+-- with the online solver process
+withPredomain ::
```
travitch, line 219: We should probably document what `withPredomain` means. What does this introduce into the assumption environment?

src/Pate/Verification/AbstractDomain.hs
```diff
@@ -0,0 +1,506 @@
...
+      }
+      -> MacawAbstractValue sym tp
+  deriving Eq
+
+-- | An abstract value for a pointer, with a separate 'AbsRange' for the region and offset.
+data MemAbstractValue sym w where
+  MemAbstractValue :: (MacawAbstractValue sym (MT.BVType (8 W4.* w))) -> MemAbstractValue sym w
+
+
+-- | Mapping from locations to abstract values.
```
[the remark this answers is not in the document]
danmatichuk: At the moment we aren't reasoning at all about this, no. This is establishing constraints about values sitting in memory, but doesn't have any information about the pointers themselves. I think this would make sense as a separate kind of abstract domain, which specifically tracks information about symbolic pointers.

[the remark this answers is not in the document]
travitch: Can we just add a note about that and some potential issues? We don't need to fix it right now

src/Pate/Verification/AbstractDomain.hs
danmatichuk: Yes, I'll add comments.

src/Pate/Verification/AbstractDomain.hs
danmatichuk: Yes, I'll add comments.

src/Pate/Verification/Widening.hs
danmatichuk: Added a comment.

PR #290, opened 2022-05-20
Update the macaw submodule to bring support for conditional calls

### LINKED ISSUE #284: Support ARM conditional calls
We now have support for ARM conditional returns via a macaw architecture-specific extension. We need to add the analogous infrastructure for conditional calls.

Contributes to #282

travitch:
This adds support for conditional calls, which are now decoded and lifted into Crucible correctly. Note that we still need to update the verifier to handle them properly.


travitch:
@danmatichuk This start of a PR lays the groundwork for conditional calls. I think you have more context on how the pate-level slicing needs to be updated to deal with the new block terminators. Could you take a look at that when you have a chance?

Issue #295, opened 2022-06-03
Unexpected equivalence proof when assigning from an unobservable global


danmatichuk:
The strongest postcondition verifier manages to prove these programs are equivalent:

Given the same preamble:
```
int h = -11;
int g __attribute__((section(".output"))) = -12;

void f(int a){
  h = a;
}
```

Where `h` is an unobservable global and `g` is observable.

And these two test functions:

Original:
```
void test(){
  f(1);
  g = h;
}
```

Patched:
```
void test(){
  f(2);
  g = h;
}
```

The verifier claims they are observably equivalent.

danmatichuk:
The same issue arises if the write occurs outside of the function:

```
void f(){
  return;
}

void test(){
  h = 1;
  f();
  g = h;
}
```

vs

```
void test(){
  h = 2;
  f();
  g = h;
}
```

PR #296, opened 2022-06-06
check equality for addresses in pre-domain

danmatichuk:
In the strongest postcondition verifier it is possible for
the incoming pre-domain to relax equality on locations that
the slice does not write to.

When computing the predicate to test for equality, we therefore
need to additionally check whether or not these locations are equal,
otherwise we will unsoundly omit them from the resulting domain
and implicitly assume that they are now equal.

Fixes: #295


src/Pate/Equivalence.hs
```diff
@@ -326,11 +327,12 @@ memDomPost ::
   MemRegionEquality sym arch ->
   SimOutput sym arch PBi.Original ->
   SimOutput sym arch PBi.Patched ->
-  PEM.MemoryDomain sym arch ->
+  PEM.MemoryDomain sym arch {- ^ pre-domain for this slice -} ->
+  PEM.MemoryDomain sym arch {- ^ target post-domain -} ->
```
travitch, line 331: The haddock for this function refers to "the given 'PEM.MemPred'". It now takes two `PEM.MemoryDomain` instead. Can you update the comment to reflect that? This change seems slightly subtle, so we should capture what the semantics are here (and it is now important that it gets both a pre and post).

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -145,6 +146,18 @@ mux sym p predT predF = case W4.asConstantPred p of
     locs <- PMC.muxMemCellPred sym p (memDomainPred predT) (memDomainPred predF)
     return $ MemoryDomain locs pol
 
+-- | Convert a domain into a negative polarity domain.
+-- If it is already a negative domain, this is a no-op. Otherwise
+-- the result is the universal domain.
```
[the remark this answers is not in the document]
danmatichuk: I've removed this from the MemoryDomain module and instead am just flipping the domain manually in the one place we need it. It's technically addressing a case that shouldn't actually occur, so really it's just technical debt from supporting polarities. I left a note explaining why it's necessary for the moment, but ultimately we should be dropping the memory domain polarities once we deprecate the backwards solver.

src/Pate/Equivalence.hs
danmatichuk: Added some comments explaining what each given MemoryDomain is used for

travitch (approved):

Issue #204, opened 2022-03-09
Factor out read/write filtering in `fromFootprints`


danmatichuk:
In `MemoryDomain.fromFootprints` the set of memory footprints is filtered according to the polarity predicate given: positive polarity only includes Read operations while negative polarity only include Write operations.

This is conflating two operations, since this is only utilized in `Verification.Domain.guessMemoryDomain` as part of the pre-domain propagation.

PR #303, opened 2022-06-23
remove polarity from memory domain

### LINKED ISSUE #276: Remove polarity from memory domains
With the forward (strongest postcondition) verifier, the polarity of all memory domains is always negative (i.e. memory domains only accumulate values that are not equivalent). By committing to only supporting negative-polarity domains, we can significantly simplify their implementation.

danmatichuk:
fixes #276 and #204

### DIFF
--- a/src/Pate/Equivalence/MemoryDomain.hs
+++ b/src/Pate/Equivalence/MemoryDomain.hs
@@ -19,7 +19,7 @@ module Pate.Equivalence.MemoryDomain (
   , cells
   , fromFootPrints
   , excludeFootPrints
-  , containsCell
+  , mayContainCell
   , mux
   , intersect
   , ppMemoryDomainEntries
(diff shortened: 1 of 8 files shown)


src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -49,26 +49,23 @@ import qualified Pate.Parallel as Par
...
+-- all addresses are considered to be "in" the domain unless they are
+-- present in the given 'PMC.MemCellPred'.
+-- Notably this does not consider the memory region covered by a
+-- given 'PMC.MemCell'. A cell must be exactly included in the
+-- underlying 'PMC.MemCellPred' to be excluded by the domain -
+-- it is not sufficient for a cell to be subsumed by an entry.
 --
--- The interpretation of those predicates is subject to the 'memDomainPolarity'.
+-- This is sound, less precise, and significantly easier to
+-- prove membership.
```
travitch, line 61: What is it less precise than, or significantly easier to prove membership compared to?

travitch (approved):
Love it. Just had one clarification request in there

src/Pate/Equivalence/MemoryDomain.hs
```diff
@@ -156,68 +159,44 @@ containsCell ::
...
+universal _sym = MemoryDomain (PMC.MemCellPred M.empty)
 
 -- | Derive a 'MemoryDomain' from a set of 'MT.MemFootprint'.
--- The semantics of this domain depend on the given polarity: for positive polarity
--- the resulting domain includes all of the given footprints. For negative polarity,
--- the resulting domain includes all of memory, except for the given footprints.
--- TODO: This also filters the set of footprints according to the given polarity.
--- For positive polarity, the footprints are filtered to only contain the reads,
--- while negative polarity filters only the writes. This should be factored into
--- a separate step (see https://github.com/GaloisInc/pate/issues/204)
```
travitch, line 177: Does this close #204 also?

src/Pate/Equivalence/MemoryDomain.hs
danmatichuk: Good question. I moved this discussion to the function where it is relevant and expanded on it a bit.

PR #305, opened 2022-06-30
More precise handling of stack

### LINKED ISSUE #278: More precise handling of stack
The SP verifier currently has a pretty weak notion of stack handling. (details to come).

danmatichuk:
This improves our handling of stack offsets by ensuring that stack accesses in the equivalence domain are always represented with respect to a fixed FRAME parameter (representing the base stack value at the function entry point), allowing the domains to be interpreted consistently across slices.

Currently our stack analysis is very weak, as stack pointers are always free variables at the start of any slice.

```
void f() { return; // does nothing }

void test() {
  // slice 0
  // SP := SP_0
  // domain := {}
  int i = 1;
  int j = 2 OR 3;  // 2 in original function, 3 in patched function
  // SP := SP_0 - 2
  // domain := { SP_0 - 1 }
  // end of slice 0
  f();
  // slice 1
  // SP := SP_1
  // domain := { SP_0 - 1}
  g = i;
  // end of slice 1
}
```

As a result of this weakness, we are unable to verify equality for this simple function as we cannot determine that `i` after the `f()` returns is the same `i` before the call.

Now the FRAME parameter ensures that the equivalence domain is meaningful in the slice at the return site for `f()`.

```
void test() {
  // SP := FRAME
  // domain := {}
  int i = 1;
  int j = 2 OR 3;
  // SP := FRAME - 2
  // domain := { FRAME - 1 }
  f();
  // SP := FRAME - 2
  // domain := { FRAME - 1}
  g = i;
}
```

fixes #278


src/Pate/Monad.hs
```diff
@@ -423,8 +429,27 @@ withAssumptionSet asm f = withSym $ \sym -> withSolverProcess $ \sp -> do
   CMR.local (\env -> env { envCurrentFrame = Some (asm <> curAsm) }) $ do
     IO.withRunInIO $ \inIO -> WPO.inNewFrame sp $ do
       W4.assume (WPO.solverConn sp) p
+      inIO $ validateAssumptions curAsm asm
       inIO f
 
+validateAssumptions ::
```
travitch, line 435: Is it the case that this is validating that the assumptions that are active in the solver are consistent (i.e., not inherently contradictory)? If so, I think it is worth noting that the `AssumptionSet`s passed in are *not* what is actually getting checked -- rather, they seem to be for error messages.

src/Pate/Monad.hs
```diff
@@ -423,8 +429,27 @@ withAssumptionSet asm f = withSym $ \sym -> withSolverProcess $ \sp -> do
...
+  AssumptionSet sym v ->
+  AssumptionSet sym v ->
+  EquivM sym arch ()
+validateAssumptions oldAsm newAsm = withSym $ \sym -> withSolverProcess $ \sp -> IO.withRunInIO $ \inIO -> do
+  let
+    simp :: forall tp. W4.SymExpr sym tp -> IO (W4.SymExpr sym tp)
+    simp e = resolveConcreteLookups sym (\e1 e2 -> W4.asConstantPred <$> liftIO (W4.isEq sym e1 e2)) e  >>= simplifyBVOps sym >>= expandMuxEquality sym
+  
+  oldAsm' <- liftIO $ PEM.mapExpr sym simp oldAsm
+  newAsm' <- liftIO $ PEM.mapExpr sym simp newAsm
```
travitch, line 447: Presumably these simplifications are solely for readability?

src/Pate/Monad.hs
```diff
@@ -459,19 +484,27 @@ applyAssumptionSet asm f = withSym $ \sym -> do
...
+  p <- liftIO $ getAssumedPred sym asm
+  case W4.asConstantPred p of
+    Just False -> return Nothing
+    Just True -> Just <$> f
+    _ -> do
+      curAsm <- currentAsm
+      CMR.local (\env -> env { envCurrentFrame = Some (asm <> curAsm) }) $ do
+        IO.withRunInIO $ \inIO -> WPO.inNewFrame sp $ do
+          W4.assume (WPO.solverConn sp) p
+          b <- WPO.checkAndGetModel sp "check assumptions" >>= asSat
```
travitch, line 504: Is there some argument we can write down for why we are not concerned that we haven't specified a timeout here? Is this always easy enough to succeed? Can we not continue if we don't have an accurate answer?

src/Pate/Verification/PairGraph.hs
```diff
@@ -300,18 +300,21 @@ reportAnalysisErrors logAction gr =
...
+
+
+
 initialDomainSpec ::
   forall sym arch.
   GraphNode arch ->
   EquivM sym arch (PAD.AbstractDomainSpec sym arch)
 initialDomainSpec (GraphNode blocks) = withFreshVars blocks $ \_vars -> do
-  dom <- initialDomain
+  dom <- initialDomain
```
travitch, line 311:
```suggestion
  dom <- initialDomain
```

src/Pate/Verification/SymbolicExecution.hs
```diff
@@ -255,8 +255,12 @@ simulate simInput = withBinary @bin $ do
   globals <- getGlobals simInput
   cres <- evalCFG globals regs cfg
   (asm, postRegs, memTrace, exitClass) <- getGPValueAndTrace cres
-
-  return $ (asm, PS.SimOutput (PS.SimState memTrace postRegs) exitClass)
+  -- in general we don't know anything about the post-state frame
+  -- we need additional assumptions based on the exit condition of this
+  -- slice
```
travitch, line 260: Are those additional assumptions introduced here? Are they handled somewhere else?

src/Pate/Verification/Widening.hs
```diff
@@ -171,31 +179,123 @@ widenAlongEdge bundle from d gr to = withSym $ \sym ->
...
 abstractOverVars ::
-  SimBundle sym arch v ->
+  forall sym arch pre.
+  PS.SimScope sym arch pre  ->
+  SimBundle sym arch pre ->
   GraphNode arch {- ^ source node -} ->
   GraphNode arch {- ^ target graph node -} ->
   PAD.AbstractDomainSpec sym arch {- ^ previous post-domain -} ->
-  PAD.AbstractDomain sym arch v {- ^ computed post-domain -} ->
+  PAD.AbstractDomain sym arch pre {- ^ computed post-domain -} ->
```
travitch, line 214: Is this the postdomain with (unsafe) references to variables in the predomain, such that the function fixes up the references to be correct in the returned `AbstractDomainSpec`? I think that interpretation makes sense, but it was a bit confusing to see this argument commented as a post-domain while seeing `pre` in its type.

I think this could just be clarified by tweaking the comment on this argument to say something like `computed post-domain (with unsafe references to pre-domain variables)`

src/Pate/Verification/Widening.hs
```diff
@@ -171,31 +179,123 @@ widenAlongEdge bundle from d gr to = withSym $ \sym ->
...
+-- | Compute an'PAD.AbstractDomainSpec' from the input 'PAD.AbstractDomain' that is
+-- parameterized over the *output* state of the given 'SimBundle'.
+-- Until now, the widening process has used the same scope for the pre-domain and
+-- post-domain (i.e. both contain free variables corresponding to the initial values
+-- before symbolic execution).
+-- To abstract the computed post-domain from its calling context, we need to rephrase
+-- any symbolic terms it contains so that they only refer to the output state.
+-- Specifically, given a post-domain @dom[pre]@ phrased over the pre-state variables, and
+-- a function @f(pre)@ representing the result of symbolic execution, we want to compute
+-- @dom'[post]@ such that @dom'[post/f(pre)] == dom[pre]@.
```
travitch, line 206: Can you add an explanation here that there are multiple strategies for accomplishing this (i.e., that list of `Alternative`s at the end)? I don't think you need to explain each one in detail, but that there are multiple and at least one is expected to work for each case.

src/Pate/Verification/Widening.hs
```diff
@@ -171,31 +179,123 @@ widenAlongEdge bundle from d gr to = withSym $ \sym ->
...
+        <|> asSimpleAssign se
+      case mres of
+        Just se' -> return se'
+        -- FIXME: add graceful failure mode
+        Nothing -> do
+          se' <- liftIO $ PS.applyScopeCoercion sym pre_to_post se
+          e'' <- liftIO $ PS.applyScopeCoercion sym post_to_pre se'
+          curAsms <- currentAsm
+
+          fail $ "Unable to rescope:" ++ (show curAsms) ++ (show (W4.printSymExpr (PS.unSE se))) ++ "***" ++ (show (W4.printSymExpr (PS.unSE e'')))
```
travitch, line 298: Is it worth just making an exception for this now?

src/What4/ExprHelpers.hs
```diff
@@ -294,8 +295,12 @@ rewriteSubExprs'' sym cache f e_outer = do
   -- This ensures that the What4 expression builder correctly re-establishes any invariants it requires
   -- after rebinding.
   Pair vars vals <- return $ toBindings binds
-  fn <- W4.definedFn sym W4.emptySymbol vars e' W4.AlwaysUnfold
-  W4.applySymFn sym fn vals >>= fixMux sym
+  case Ctx.viewSize (Ctx.size vals) of
+    -- no replacement
```
travitch, line 299: It seems like this is a subtle change - can you document what would go wrong if you didn't special case this? I'm assuming this was our sneaky source of unsoundness

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -877,5 +962,7 @@ mkSimBundle pPair vars d =
 
      let bnd = SimBundle (PPa.PatchPair simInO simInP) (PPa.PatchPair simOutO_ simOutP_)
 
-     eqRegs <- PVD.equateRegisters rd bnd
-     withAssumptionSet eqRegs $ applyCurrentAsms bnd
+     applyCurrentAsms bnd
+
+     -- eqRegs <- PVD.equateRegisters rd bnd
+     -- withAssumptionSet eqRegs $ applyCurrentAsms bnd
```
travitch, line 968: Any reason to not just delete this?

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -223,38 +271,63 @@ visitNode (ReturnNode fPair) vars d gr0 =
...
 --   information from function return nodes to the actual return sites.
+--   The only state update that occurs is that a fresh variable is created
+--   to represent the stack base in the output state.
+--   The returned 'PS.AssumptionSet' associates the internal stack base
+--   from the callee (i.e. in 'SimInput') with the stack
+--   pointer of the caller (i.e. in 'SimOutput').
+--   This effectively re-phrases the resulting domain from the
+--   function call with respect to the specific call site that we are
+--   returning, and allows us to properly contextualize any resulting inequalities
+--   with respect to what is visible in our stack frame.
```
travitch, line 297: Is this making a strong non-interference assumption about the callee? Or is that what is explicitly being propagated to the caller here?

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -157,11 +157,57 @@ pairGraphComputeFixpoint gr =
...
+
+absValueToAsm ::
+  forall sym arch v bin tp.
+  PS.SimVars sym arch v bin ->
+  PSR.MacawRegEntry sym tp ->
+  MAS.AbsValue (MM.ArchAddrWidth arch) tp ->
+  EquivM sym arch (PS.AssumptionSet sym v)
+absValueToAsm vars regEntry val = withSym $ \sym -> case val of
+  -- FIXME: check the MemAddr here to make sure we only use
+  -- stack offsets from this frame
```
travitch, line 175: Which `MemAddr` is this referring to? Is that the `_` in the pattern match on the `StackOffsetAbsVal`?

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -157,11 +157,57 @@ pairGraphComputeFixpoint gr =
...
+          PS.exprBinding (W4.natToIntegerPure region) (W4.natToIntegerPure stackRegion)
+
+    let w = MM.memWidthNatRepr @(MM.ArchAddrWidth arch)
+    slotBV <- liftIO $ W4.bvLit sym w (BV.mkBV w (fromIntegral slot))
+    let sb = PS.unSE $ PS.simStackBase $ PS.simVarState vars
+    off' <- liftIO $ W4.bvAdd sym sb slotBV
+    -- the offset of this value must be frame + slot
+    let bindOffSet = PS.exprBinding off off'
+    return $ bindRegion <> bindOffSet
+  _ -> return $ mempty
```
travitch, line 190:
```suggestion
  _ -> return mempty
```

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -157,11 +157,57 @@ pairGraphComputeFixpoint gr =
...
-        withAssumptionSet asm $
-          visitNode nd vars d gr'
+      gr'' <- PS.viewSpec preSpec $ \scope d -> do
+        withAssumptionSet (PS.scopeAsm scope) $
+          visitNode scope nd d gr'
       pairGraphComputeFixpoint gr''
 
+
+
+absValueToAsm ::
```
travitch, line 167: How does the `AbsValue` turn into an `AssumptionSet`? It looks like, at a high level, this is doing the offset remapping step. Probably worth a few words here, or at least a pointer to the overall description of the tricky bits here

src/Pate/Verification/AbstractDomain.hs
```diff
@@ -392,16 +394,25 @@ applyAbsRange sym e rng = case rng of
 absDomainValToAsm ::
   W4.IsSymExprBuilder sym =>
   sym ->
+  PE.EquivContext sym arch ->
   PSR.MacawRegEntry sym tp ->
+  Maybe (MAS.AbsValue (MM.ArchAddrWidth arch) tp) ->
   MacawAbstractValue sym tp ->
```
travitch, line 400: There are now two abstract values here - can you add some commentary on the difference between the two?

src/Pate/SimState.hs
```diff
@@ -162,6 +195,36 @@ instance OrdF (W4.SymExpr sym) => Semigroup (AssumptionSet sym v) where
...
     in AssumptionSet preds binds
 
+ppExprSet ::
+  W4.IsExpr (W4.SymExpr sym) =>
+  Proxy sym ->
+  ExprSet sym tp ->
+  PP.Doc a
+ppExprSet _ es =
+  let ps = [ W4.printSymExpr p | p <- SetF.toList es ]
+  in PP.sep (zipWith (<+>) ("{" : repeat ",") ps) <+> "}"
```
travitch, line 205: Any reason this isn't in `ExprHelpers` with the definition of `ExprSet`?

src/Pate/SimState.hs
```diff
@@ -75,49 +89,68 @@ import           GHC.Stack ( HasCallStack )
...
 import qualified Data.Parameterized.SetF as SetF
 
 ------------------------------------
 -- Crucible inputs and outputs
 
-
+-- | Points to the base of the stack. In any given context this will always be
+-- "free" as the base of the stack is always abstract, but it is rebound to account
+-- for changes to the stack pointer when moving between scopes.
+type StackBase sym arch v = ScopedExpr sym (W4.BaseBVType (MM.ArchAddrWidth arch)) v
```
travitch, line 136: My usual annoying question: if we want to name this as something special, does it deserve a `newtype`?

src/Pate/SimState.hs
```diff
@@ -490,62 +553,185 @@ mkVarBinds ::
...
+  forall sym t solver fs v v'.
+  sym ~ (W4B.ExprBuilder t solver fs) =>
+  ExprRewrite sym v v' ->
+  IO (ScopeCoercion sym v v')
+asScopeCoercion rew = ScopeCoercion <$> freshVarBindCache <*> pure rew
+
+-- | An expr tagged with a scoped parameter (representing the fact that the
+-- expression is valid under the scope 'v')
+data ScopedExpr sym tp (v :: VarScope) =
+  ScopedExpr { unSE :: (W4.SymExpr sym tp) }
```
travitch, line 622:
```suggestion
  ScopedExpr { unSE :: W4.SymExpr sym tp }
```

travitch:
Also, could you add something like a
```
{- Note [Stack Relative References]

-}
```

that discusses the high-level design of this feature that is referenced from the places it is implemented? The discussion could largely be taken from that Google Doc you already wrote. It would be very helpful to have that near the code, as it isn't obvious, without already knowing what the problems are, why this is all happening.

src/Pate/Monad.hs
danmatichuk: Added a comment

src/Pate/Monad.hs
danmatichuk: Yes. I'll put a comment in, but in general it would be good to have this kind of presentation-specific simplification logic exist in one place.

danmatichuk: https://github.com/GaloisInc/pate/issues/315

src/Pate/Monad.hs
danmatichuk: I've got a patch that addresses this that will be committed later. Specifically I had to put in some fixes to get timeouts working again: https://github.com/GaloisInc/pate/issues/316

I'll put a comment here.

src/Pate/Verification/SymbolicExecution.hs
danmatichuk: No they are introduced later, I added a comment with some signposting

src/Pate/Verification/Widening.hs
danmatichuk: Yes that's about right. The resulting 'AbstractDomainSpec' is scope-agnostic (i.e. the body of the spec should only be scoped to the variables from that spec), so we have removed references to the pre-scoped variables in the result.

I added your suggested comment, I'm not sure if this needs to be clarified further though.

src/Pate/Verification/Widening.hs
danmatichuk: Added comment

src/Pate/Verification/Widening.hs
danmatichuk: I have an incoming patch that handles this more carefully by emitting warning and dropping the element from the domain, so this 'fail' will go away once that is in place.

src/What4/ExprHelpers.hs
danmatichuk: It wasn't a source of unsoundness, but I suspected that it was trashing expression nonces in some cases and causing cache misses that would otherwise hit. I can drop the change from this PR if we want, but it shouldn't actually affect any semantics.

travitch: No reason not to include it - I just like to understand subtle changes to avoid surprise.

src/Pate/Verification/StrongestPosts.hs
danmatichuk: No this isn't really making any assumptions about the callee, since any changes it has made (that result in inequalities in the domain) will appear in the incoming domain (i.e. the domain passed to 'visitNode').

Basically this is a "virtual" transition where all we are doing is taking the outgoing domain from the callee and re-phrasing it to be based off of the stack base of the caller. So yes, any changes will still be observable to the caller (assuming 'abstractOverVars' was successful during this widening step).

src/Pate/Verification/StrongestPosts.hs
danmatichuk: Yes, I'll bind it to something just so it's not mysterious.

src/Pate/Verification/StrongestPosts.hs
danmatichuk: Added a comment

src/Pate/Verification/AbstractDomain.hs
danmatichuk: Added comment

src/Pate/SimState.hs
danmatichuk: Moved

src/Pate/SimState.hs
danmatichuk: Agreed, I've wrapped it in a `newtype` now

travitch (approved):

PR #309, opened 2022-07-08
Clean up solver usage and assumption tracking

### LINKED ISSUE #89: Clean up logic operations in EquivM
The `EquivM` monad defines a collection of logic operations for handling assumptions, proving simple predicates, and managing the scoping of variables. Ideally, this could be either abstracted or at least made more structured. For example, `withAssumption` has several variants depending on wether or not the inner function may yield additional assumptions, but it's cumbersome to keep track of which variant should be used in which situation.

### LINKED ISSUE #301: Add abstraction over solver usage in strongest postcondition verifier
The strongest postcondition veriifier assumes that an online what4 solver is available, but each invocation of the solver involves directly calling `Lang.Crucible.Backend.Online.withSolverProcess` and managing assumptions using `What4.Protocol.Online.assume`.

Most of these invocations occur inside the `EquivM` monad, so we should wrap this in an abstraction that removes a lot of this boilerplate and consistently handles assumptions.

### LINKED ISSUE #307: Enforce abstraction boundary for SimSpec
Currently SimSpec exposes a fairly leaky abstraction - we should be able to restrict its interface to avoid exposing its internal structure. In particular we should be able to have wrapper functions for cleanly producing a fresh SimSpec.

danmatichuk:
This cleans up some of the logic operations surrounding assumptions by assuming that an online solver is always available, and using its concept of assumptions/frames.

It also renames AssumptionFrame to AssumptionSet and adds a scope parameter to it, and changes SimSpec to use this as its internal assumption storage.

fixes #89 
fixes #307 
fixes #301


src/Pate/Monad.hs
```diff
@@ -380,126 +388,96 @@ freshSimVars blocks = do
   regs <- MM.mkRegStateM (\r -> unconstrainedRegister argNames r)
   return $ SimBoundVars regs (SimState mem (MM.mapRegsWith (\_ -> PSR.macawVarEntry) regs))
 
-
+-- Although 'AssumptionSet' has a scope parameter, the current interface doesn't have a
+-- good mechanism for enforcing the fact that we are only pushing assumptions that
+-- are actually scoped properly.
+-- This is manifest in the fact that the background frame in 'envCurrentFrame' doesn't
+-- track any scope, and is therefore unsafely coerced into any target scope.
+-- TODO: Rather than trying to enforce this statically (which would be difficult and require
```
travitch, line 396: This seems important to not lose track of - can we add a github issue that we can reference from the code here?

src/Pate/SimState.hs
```diff
@@ -212,35 +243,38 @@ macawRegBinding sym var val = do
     CT.BoolRepr -> return $ exprBinding (PSR.macawRegValue var) (PSR.macawRegValue val)
     _ -> return mempty
 
+-- TODO: this is generally unsafe, since we don't check that the incoming
```
travitch, line 246: Can you note what would go wrong here if this became a problem?

travitch (approved):
Much nicer - can you clean up some of the warnings on unused imports and some newly unused variables that were introduced?

PR #314, opened 2022-07-25
Add better simplifier tracing and fix unsound simplification step

### LINKED ISSUE #141: Debug simplifier failures
The deep formula simplification pass has been observed to (rarely) generate simplified terms not equivalent to the originals. It has been instrumented to report when this happens (via Debug.Trace). We need to figure out which simplification rule is at fault.

danmatichuk:
fixes #141

### DIFF
--- a/src/Pate/Verification/Simplify.hs
+++ b/src/Pate/Verification/Simplify.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 module Pate.Verification.Simplify (
     simplifyPred
   , simplifySubPreds
@@ -12,6 +13,7 @@ module Pate.Verification.Simplify (
 import           Control.Monad.IO.Class ( liftIO )
 import qualified Control.Monad.IO.Unlift as IO
 import qualified Control.Monad.Reader as CMR
+import           Data.Proxy
 import           Data.Functor.Const ( Const(..) )
 import           Debug.Trace ( traceM )
 import           GHC.Stack ( HasCallStack )
@@ -21,6 +23,7 @@ import qualified What4.SatResult as W4R
 
 import qualified Pate.Config as PC
 import qualified Pate.ExprMappable as PEM
+import qualified Pate.Equivalence.Error as PEE
 import           Pate.Monad
 import qualified What4.ExprHelpers as WEH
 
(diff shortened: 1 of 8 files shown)


src/What4/ExprHelpers.hs
```diff
@@ -973,36 +1005,72 @@ simplifyBVOpInner sym go app = case app of
...
+    f app = case simplifyBVOpInner sym (\e1 e2 -> inIO (simp_check e1 e2)) (\e' -> inIO (go e')) app of
+      Just g -> Just <$> liftIO g
+      Nothing -> return Nothing
+
+    go :: forall tp'. W4.SymExpr sym tp' -> m (W4.SymExpr sym tp')
+    go e = W4B.idxCacheEval cache e $ simplifyApp sym simp_check f e
+
+  inIO (go outer)
+
+simplifyApp ::
```
travitch, line 1038: What are the expected uses/semantics of the "double-check simplification step" - is this a full simplification validation?

src/Pate/Verification/Simplify.hs
```diff
@@ -54,6 +57,47 @@ simplifySubPreds a = withValid $ withSym $ \sym -> do
...
+  W4.SymExpr sym tp ->
+  EquivM sym arch (W4.SymExpr sym tp)
+simplifyBVOps_trace sym checkWork outer = do
+  cache <- W4B.newIdxCache
+  let
+    go :: forall tp'. W4.SymExpr sym tp' -> EquivM_ sym arch (W4.SymExpr sym tp')
+    go e = W4B.idxCacheEval cache e $ WEH.simplifyBVOps' sym checkWork e
+  go outer
+
+simplifyWithSolver ::
```
travitch, line 78: What does this use the solver for? What is the behavior if there is a timeout?

src/Pate/Verification/Simplify.hs
```diff
@@ -3,10 +3,13 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
```
travitch, line 6: Is this extension actually used? I didn't see any type applications at a glance

src/What4/ExprHelpers.hs
danmatichuk: I wrapped up this check in a newtype and added some comments

src/Pate/Verification/Simplify.hs
danmatichuk: Added a comment explaining this

src/Pate/Verification/Simplify.hs
danmatichuk: Removed

travitch (approved):
Thanks, looks good

PR #317, opened 2022-07-27
Handle timeouts when using the online solver

danmatichuk:
This adds some error handling surrounding the use of the `timeout` primitive for querying the solver, since we need to do some cleanup when the underlying solver process is killed.

### DIFF
--- a/src/Pate/Monad.hs
+++ b/src/Pate/Monad.hs
@@ -609,11 +609,11 @@ checkSatisfiableWithModel ::
   EquivM sym arch (Either SomeException a)
 checkSatisfiableWithModel timeout desc p k = withSym $ \sym -> do
   st <- withOnlineBackend $ \bak -> liftIO $ LCB.saveAssumptionState bak
-  mres <- withSolverProcess $ \sp -> IO.withRunInIO $ \runInIO -> do
+  mres <- withSolverProcess $ \sp -> liftIO $ do
     WPO.push sp
     W4.assume (WPO.solverConn sp) p
     tryJust filterAsync $ do
-      res <- runInIO (checkSatisfiableWithoutBindings timeout sym desc $ WPO.checkAndGetModel sp "checkSatisfiableWithModel")
+      res <- checkSatisfiableWithoutBindings timeout sym desc $ WPO.checkAndGetModel sp "checkSatisfiableWithModel"
       W4R.traverseSatResult (\r' -> pure $ SymGroundEvalFn r') pure res
   case mres of
     Left err -> withOnlineBackend $ \bak -> do
@@ -644,12 +644,12 @@ checkSatisfiableWithoutBindings
   -> sym
   -> String
   -> IO (W4R.SatResult (W4G.GroundEvalFn t) ())
-  -> EquivM sym arch (W4R.SatResult (W4G.GroundEvalFn t) ())
+  -> IO (W4R.SatResult (W4G.GroundEvalFn t) ())
 checkSatisfiableWithoutBindings timeout _sym _desc doCheckSat =
   case PT.timeoutAsMicros timeout of
-    Nothing -> liftIO $ doCheckSat
+    Nothing -> doCheckSat
     Just micros -> do
-      mres <- liftIO $ PT.timeout micros doCheckSat
+      mres <- PT.timeout micros doCheckSat
       case mres of
         Nothing -> return W4R.Unknown
         Just r -> return r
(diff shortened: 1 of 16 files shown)


src/Pate/Monad.hs
```diff
@@ -785,8 +862,34 @@ throwHere err = withValid $ do
...
+        False -> throwError er
+      ContinueAfterFailure -> return r
+    r -> return r
+
+internalizeError :: Either (PEE.EquivalenceError arch) a -> EquivM_ sym arch a
+internalizeError merr = case merr of
+  Left err -> throwError err
+  Right a -> return a
+
+-- | Run an IO operation, internalizing any exceptions raised
```
travitch, line 881: What does "internalize" mean here?

src/Pate/Verification.hs
```diff
@@ -48,6 +48,8 @@ import           Prelude hiding ( fail )
 import qualified What4.Expr as WE
 import qualified What4.FunctionName as WF
 import qualified What4.Interface as W4
+import           System.IO
+import           Data.IORef
```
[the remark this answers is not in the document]
danmatichuk: Fixed

src/Pate/Monad.hs
```diff
@@ -420,16 +426,31 @@ withFreshVars blocks f = do
...
-      inIO $ validateAssumptions curAsm asm
-      inIO f
+    (frame, st) <- withOnlineBackend $ \bak -> liftIO $ do
+      st <- LCB.saveAssumptionState bak
+      frame <- LCB.pushAssumptionFrame bak
+      LCB.addAssumption bak (LCB.GenericAssumption initializationLoc "withAssumptionSet" p)
+      return (frame, st)
+    mr <- manifestError (f >>= \r -> validateAssumptions curAsm asm >> return r)
+    safePop frame st
+    internalizeError mr
```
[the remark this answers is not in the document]
danmatichuk: They can be in the same `withOnlineBackend` scope, I can easily refactor that in. The problem I was finding when using `inNewFrame` was that it misbehaved when the underlying solver process is killed by the timeout. All of this extra bracketing is to make sure that the verifier can continue in a consistent state even if a timeout is tripped or the solver is killed.

src/Pate/Monad.hs
danmatichuk: It's essentially the inverse of `manifestError` - raise an error if it exists and otherwise return. I'm happy to change it to either `unManifestError` or `raiseIfError`, whichever makes more sense.

src/Pate/Monad/Environment.hs
```diff
@@ -52,6 +52,7 @@ import qualified Pate.Verification.Override.Library as PVOL
 data VerificationFailureMode =
     ThrowOnAnyFailure
   | ContinueAfterFailure
```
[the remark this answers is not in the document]
danmatichuk: Not at the moment. I need to add a flag to set this from the command line, but just haven't done it yet.

src/Pate/Monad.hs
travitch: I think those are all still confusing (except maybe `raiseIfError`). The slightly non-obvious thing that I think is worth capturing is that this turns Haskell IO exceptions into (slightly cleaner?) `ExceptT` exceptions in `EquivM_`.

I suppose that is the definition of "internalize"

src/Pate/Monad/Environment.hs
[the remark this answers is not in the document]
travitch: Gotcha

src/Pate/Monad.hs
danmatichuk: I've removed it in favor of using the more standard 'finalize' function from Monad.Catch, which I think is more clear.

src/Pate/Monad.hs
```diff
@@ -555,13 +642,14 @@ checkSatisfiableWithoutBindings
   :: (sym ~ WE.ExprBuilder t st fs)
   => PT.Timeout
   -> sym
+  -> String
   -> IO (W4R.SatResult (W4G.GroundEvalFn t) ())
-  -> IO (W4R.SatResult (W4G.GroundEvalFn t) ())
-checkSatisfiableWithoutBindings timeout _sym doCheckSat =
+  -> EquivM sym arch (W4R.SatResult (W4G.GroundEvalFn t) ())
```
travitch, line 647: I just noticed that this changed from `IO` to `EquivM`, but it doesn't look like anything in here uses that additional context?

Is this intentional or a forward looking change? I'm not opposed, but curious.

travitch (approved):

src/Pate/Monad.hs
danmatichuk: No I think it made sense at some point but eventually all of the EquivM stuff eventually ended up outside the function. I'll just switch it back to IO for clarity.

PR #319, opened 2022-07-29
Various optimizations

### LINKED ISSUE #313: Avoid exponential case when verifying observable traces
The function `StrongestPosts.equivalentSequences` computes a predicate that is true iff the two input sequences (of memory events) are equivalent. There is a potential exponential blowup in the case that handles `SymSequenceMerge`: in the general case we need to split the entire analysis over the predicate that the sequence is merged on. In the simple case, the head of both sequences is a merge over the same predicate, and the analysis doesn't need to be forked. In the general case, this split can cause an exponential blowup.

There are two mitigations for avoiding the exponential case:

1) Use the solver to check when the merge predicates are equal (rather than relying on term equality).

2) Throw an exception (or simply report a failure) in the case where we need to split the merge on a non-trivial sequence (where non-trivial is defined as some configurable maximum sequence length).

danmatichuk:
This  is several optimizations that use caching and some changes to traversal and heuristic ordering.
(depends on #318)

  * use the same equivalence domain for discoverPairs and widening
     - deduplicate discovered pair results
  * cache intermediate results in abstractOverVars
  * traverse all locations during each widening iteration
     -  avoids re-checking earlier locations repeatedly
  * fail early on any exception during testing
  * cache intermediate results when checking sequence equality

fixes #313

### DIFF
--- a/src/Pate/Location.hs
+++ b/src/Pate/Location.hs
@@ -13,8 +13,9 @@
 {-# LANGUAGE QuantifiedConstraints  #-}
 module Pate.Location (
     Location(..)
+  , LocationWitherable(..)
   , LocationTraversable(..)
-  , SingleLocation(..)
+  , LocationPredPair(..)
   ) where
 
 import           GHC.TypeLits
@@ -23,6 +24,7 @@ import           Control.Monad.Trans.Except ( throwE, runExceptT )
 import           Control.Monad.Trans.State ( StateT, get, put, execStateT )
 import           Control.Monad.Trans ( lift )
 
+import           Data.Parameterized.Classes
 import qualified Data.Macaw.Types as MT
 import qualified Data.Macaw.CFG as MM
 
@@ -57,16 +59,27 @@ instance PEM.ExprMappable sym (Location sym arch l) where
 
 -- TODO: this can be abstracted over 'W4.Pred'
 
--- | Defines 'f' to be viewed as a traversable collection of
+-- | Defines 'f' to be viewed as a witherable (traversable but
+-- optionally dropping elements instead of updating them) collection of
 -- 'Location' elements paired with 'W4.Pred' elements.
-class LocationTraversable sym arch f where
+class LocationWitherable sym arch f where
   -- | Traverse 'f' and map each element pair, optionally dropping
   -- it by returning 'Nothing'
-  traverseLocation ::
+  witherLocation ::
     IO.MonadIO m =>
     sym ->
     f ->
     (forall l. Location sym arch l -> W4.Pred sym -> m (Maybe (Location sym arch l, W4.Pred sym))) ->
+    m f
+
+-- | Defines 'f' to be viewed as a traversable collection of
+-- 'Location' elements paired with 'W4.Pred' elements.
+class LocationTraversable sym arch f where
+  traverseLocation ::
+    IO.MonadIO m =>
+    sym ->
+    f ->
+    (forall l. Location sym arch l -> W4.Pred sym -> m (Location sym arch l, W4.Pred sym)) ->
     m f
 
   -- | Return the first location (according to the traversal order
@@ -83,7 +96,7 @@ class LocationTraversable sym arch f where
       _ <- traverseLocation sym body $ \loc v' -> do
         lift (f loc v') >>= \case
           Just a -> throwE a
-          Nothing -> return Nothing
+          Nothing -> return (loc, v')
       return ()
     case r of
       Left a -> return $ Just a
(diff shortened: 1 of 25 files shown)


src/Pate/Verification/StrongestPosts.hs
```diff
@@ -468,26 +470,68 @@ eqMemOp sym (MT.MemOp xptr xdir xcond xw xval xend) (MT.MemOp yptr ydir ycond yw
 
   | otherwise = return (W4.falsePred sym)
 
+
+data SeqTwoCache a b = SeqTwoCache (IORef (Map.Map (Maybe (Nonce GlobalNonceGenerator a), Maybe (Nonce GlobalNonceGenerator a)) b))
```
travitch, line 474: I'm pretty sure I understand what this is doing, but the name of this type bothered me for reasons I can't articulate.

Can we call it something like a `SeqPairCache`?

src/Pate/Verification/StrongestPosts.hs
```diff
@@ -468,26 +470,68 @@ eqMemOp sym (MT.MemOp xptr xdir xcond xw xval xend) (MT.MemOp yptr ydir ycond yw
...
-equivalentSequences :: forall sym ptrW.
-  (LCB.IsSymInterface sym, 1 <= ptrW) =>
+equivalentSequences :: forall sym arch ptrW.
+  (1 <= ptrW) =>
+  MM.MemWidth ptrW =>
+  SymSequence sym (MT.MemEvent sym ptrW) ->
+  SymSequence sym (MT.MemEvent sym ptrW) ->
+  EquivM sym arch (W4.Pred sym)
+equivalentSequences seq1 seq2 = withSym $ \sym -> do
+  cache <- liftIO $ newSeqTwoCache
```
travitch, line 514:
```suggestion
  cache <- liftIO newSeqTwoCache
```

src/Pate/Verification/Widening.hs
```diff
@@ -277,21 +301,40 @@ abstractOverVars scope_pre bundle _from _to postSpec postResult = withSym $ \sym
...
-        -- FIXME: add graceful failure mode
-        Nothing -> do
+      doRescope :: forall tp l. PL.Location sym arch l -> PS.ScopedExpr sym pre tp -> EquivM_ sym arch (MaybeF (PS.ScopedExpr sym post) tp)
+      doRescope _loc se = W4B.idxCacheEval cache (PS.unSE se) $ runMaybeTF $ do
+          se' <- (    asConcrete se
+                  <|> asStackOffset PBi.OriginalRepr se
+                  <|> asStackOffset PBi.PatchedRepr se
+                  <|> asScopedConst (W4.truePred sym) se
+                  <|> asSimpleAssign se
+                 )
```
travitch, line 311: You hinted that the order of these might be important - can we document the rationale, if so?

src/Pate/Verification/Widening.hs
```diff
@@ -511,6 +593,33 @@ widenValues sym evalFn bundle postD = do
...
+  let postD' = postD { PAD.absDomEq = pred' }
+  return (Widen WidenEquality (WidenLocs mempty (Set.fromList cells)) postD')
+
+widenRegs ::
+  [Some (MM.ArchReg arch)] ->
+  AbstractDomain sym arch v ->
+  EquivM sym arch (WidenResult sym arch v)
+widenRegs newRegs postD = withSym $ \sym -> do
+  let
+    regs' = foldl
```
travitch, line 614: You probably want `Data.Foldable.foldl'` here - `foldl` is nearly always a nasty space leak

src/Pate/MemCell.hs
```diff
@@ -97,6 +100,22 @@ traverseWithCell ::
 traverseWithCell (MemCellPred memPred) f =
   MemCellPred <$> Map.traverseWithKey (\(Some cell@MemCell{}) p -> f cell p) memPred
 
+rebuild ::
```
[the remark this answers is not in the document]
danmatichuk: This was just a cheap way to avoid a module import cycle, since the Location module depends on MemCell. I'll give it the same name at least so it's clear what's going on.

src/Pate/Location.hs
```diff
@@ -0,0 +1,130 @@
...
+    NoLoc -> return NoLoc
+
+
+-- TODO: this can be abstracted over 'W4.Pred'
+
+-- | Defines 'f' to be viewed as a traversable collection of
+-- 'Location' elements paired with 'W4.Pred' elements.
+class LocationTraversable sym arch f where
+  -- | Traverse 'f' and map each element pair, optionally dropping
+  -- it by returning 'Nothing'
```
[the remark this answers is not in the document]
danmatichuk: Yeah this is exactly what it is. I've added another class that clarifies traversable vs. witherable, so I think this is much cleaner now.

travitch (approved):

PR #320, opened 2022-07-29
Very simple region support in memory model

danmatichuk:
This is a very incomplete model of reading pointer regions out of memory, but supports the most common cases.

### DIFF
--- a/src/Pate/Memory/MemTrace.hs
+++ b/src/Pate/Memory/MemTrace.hs
@@ -624,7 +624,18 @@ data MemTraceState sym ptrW = MemTraceState
   }
 
 type MemTraceSeq sym ptrW = SymSequence sym (MemEvent sym ptrW)
+
+-- | A map from pointers (a region integer combined with a pointer-width bitvector)
+-- to bytes, representing the contents of memory at the given pointer.
 type MemTraceArrBytes sym ptrW = MemArrBase sym ptrW (BaseBVType 8)
+
+
+-- | A map from pointers (a region integer combined with a pointer-width bitvector)
+-- to integers, representing the region that should be used when reading a pointer
+-- back from memory (each individual byte-width slice of the pointer bitvector is assigned
+-- the region of the pointer in this map).
+-- TODO: This is a very naive model of pointer regions - there are many situations where
+-- this is not enough information to accurately recover the region of a stored pointer.
 type MemTraceArrRegions sym ptrW = MemArrBase sym ptrW BaseIntegerType
 
 type MemArrBase sym ptrW tp = RegValue sym (SymbolicArrayType (EmptyCtx ::> BaseIntegerType) (BaseArrayType (EmptyCtx ::> BaseBVType ptrW) tp))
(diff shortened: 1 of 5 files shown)


src/Pate/Memory/MemTrace.hs
```diff
@@ -649,40 +652,46 @@ initMemTrace ::
...
+  arrBytes <- ioFreshConstant sym "InitMemBytes" knownRepr
+  arrRegions <- ioFreshConstant sym "InitMemRegions" knownRepr
   sq <- nilSymSequence sym
-  return $ MemTraceImpl sq (MemTraceState arr) (toMuxTree sym Nothing) baseMem
+  return $ MemTraceImpl sq (MemTraceState arrBytes arrRegions) (toMuxTree sym Nothing) baseMem
 
 
 mkMemoryBinding ::
   forall sym ptrW.
+  IsSymExprBuilder sym =>
```
travitch, line 668: I have no objections, but there isn't anything obviously new in this function that might require the additional constraint

src/Pate/Memory/MemTrace.hs
```diff
@@ -1203,6 +1213,13 @@ readMemState :: forall sym ptrW ty.
   IO (RegValue sym (MS.ToCrucibleType ty))
 readMemState sym mem baseMem ptr repr = go 0 repr
   where
+  isPtrRead :: Bool
+  isPtrRead = case repr of
```
travitch, line 1217: Can we clarify that this is *potentially* a pointer read (because it is of pointer size) and not a guarantee that it is actually a pointer

src/Pate/Memory/MemTrace.hs
```diff
@@ -619,10 +619,13 @@ data MemTraceImpl sym ptrW = MemTraceImpl
...
+  { memArrBytes :: MemTraceArrBytes sym ptrW
+  , memArrRegions :: MemTraceArrRegions sym ptrW
+  }
 
 type MemTraceSeq sym ptrW = SymSequence sym (MemEvent sym ptrW)
-type MemTraceArr sym ptrW = MemArrBase sym ptrW (BaseBVType 8)
+type MemTraceArrBytes sym ptrW = MemArrBase sym ptrW (BaseBVType 8)
+type MemTraceArrRegions sym ptrW = MemArrBase sym ptrW BaseIntegerType
 
 type MemArrBase sym ptrW tp = RegValue sym (SymbolicArrayType (EmptyCtx ::> BaseIntegerType) (BaseArrayType (EmptyCtx ::> BaseBVType ptrW) tp))
```
travitch, line 630: Can you add a short description of how memory is arranged based on these types?  I think the basic structure can be worked out from the types, but a description would be nonetheless helpful.

More importantly, can you record the policy for choosing the region for a written pointer value?

src/Pate/Memory/MemTrace.hs
danmatichuk: I just used it to get `OrdF (SymExpr sym)`, which is needed here since we're making a nontrivial Map now.

travitch (approved):

PR #321, opened 2022-08-03
Triage challenge problems

danmatichuk:


danmatichuk:
See #326 for the continuation of this work

Issue #324, opened 2022-08-09
Test suite does not load PLT stubs from binaries


danmatichuk:
The test suites manually define the architecture proxy that is used to run the verifier, where the main tool builds the correct architecture proxy by reading the ELF headers for the given binaries. The cases the test suite to fail if the tests contain any PLT stubs, as the verifier doesn't correctly identify them.

danmatichuk:
Fixed in #326

Issue #130, opened 2021-08-25
Triage failures on challenge 3


travitch:
Inspecting the proof of challenge 3 shows a number of failures that don't have obvious explanations (and seem likely to be false positives). We should fix the easy ones and document complex cases. Note that they could relate to fixpoints (#82)

danmatichuk:
This is the current state of Challenge 3 on the solver. At the moment we are able to run the verifier up to `init_bumper` where it times out as a result of failing to infer some constant propagation.

1. Context-insensitivity is preventing constant propagation from retaining concrete region from malloc’d pointer
    * Added additional context to nodes in the PairGraph which allow each function call site to be given a distinct domain
        * Configurable at the top-level -> can be made more fine-grained
    * This is a fairly heavy-handed solution, as it is not generally clear when it should be used.
    * **TODO: Proper solution involves computing more precise “frames” for function calls.**
        * i.e. if we know a function does not modify a value, then any constraints on it at each call site can be soundly propagated across the function call
        * smells like separation logic
2. Symbolic naturals are problematic for a number of reasons
    * Added some low-level functions for extracting/injecting symbolic integers from symbolic naturals while avoiding positivity checks
        * **TODO: Needs to be applied systematically -> pointer construction/deconstruction should use a unified interface**
        * Proper solution would remove SymNat from LLVMPointer type entirely.
3. Rewrites for value/constant propagation are not applying reliably
    * When a rewrite system is non-trivial, some rewrite orderings will resolve to a constant while others will not.
        * This should not be an issue for the solver, but the terms are difficult to interpret without inlining constants.
    * Challenge is currently blocked on a problem that I believe should be trivial after simplification -> need to apply rewrites to see where the hard part is
    * Solution: **TODO: When augmenting the rewriting environment, apply the incoming rewrites to the left-hand side of all existing rewrites.**
4. Determined that the value domain propagation was failing due to improper handling of PLT stubs
    * Problem: Stack frame was not being propagated correctly
    * Solution: Updated logic for stack frame stitching to add case for handling PLT.
        * No warning mechanism for loss of precision in the value domain.
        * Suggestion: **TODO: Better top-level debugging tools that allow us to track values as they propagate**
        1. Describe the need for special case in code for PLT. 
        2. Alternatively, reuse stubs from inline callee work (based on llvm mem model - Should 2 diff mem models be unified? Then we can use same overrides for inline callee and verifier work. Avoids 2 copies of stubs) 
[...]

PR #326, opened 2022-08-25
Challenge 3 triage

### LINKED ISSUE #195: SymNats interact poorly with ExprMappable
The interface for `SymNat` maintains the invariant that its inner integer is positive by clamping its value to be necessarily positive.
This conflicts with the expression rebinding infrastructure (i.e. ExprMappable) because the only way to map an expression-modifying function over a 'SymNat' is to: extract its inner integer, apply the function, then wrap the integer back up in a SymNat.

This has a few consequences:

1) Each time a SymNat is transformed (via mapExpr) it adds a new clamp, resulting in terms like `(max 0 (max 0 (max 0 n)))`
2) Applying the identity function to a SymNat still results in a modified internal value.
3) Term equality on SymNats is unstable in this context (without term caching), since each internal int is now formally distinct.

danmatichuk:
Addressing issues that are blocking the verifier from running Challenge 3

See: #130

- Generalizes the functionality from `Main` that determines the architecture from the ELF and loads the PLT symbols from it
    - Fixes: #324 
- Include upstream What4 change that makes converting from and to SymNats less destructive to the term shape.
    - Partially fixes #195 
    - TODO: Remove workaround and update to latest what4 (see #328)
- Move `AssumptionSet` into its own module and collect its interface functions there.
   - Resolves some circular dependencies that were introduced 
   - Used in more places as the standard representation of a collection of assumptions.
   - Explicit bindings/rewrites are useful for simplifications for readability
- Use structured traversal for value domain when widening so that each value domain location is checked separately.
   - Useful for debugging which specific value constraints were causing timeouts 
- Handle cases where Crucible was throwing errors due to inconsistent assumptions
   - Introduced when solver timeout support was added. Occurs in `withSatAssumption` when the incoming assumption
      is trivially false.
- Basic PLT override support
   - Each architecture defines a mapping from PLT symbols to state transformations
   - Not intended to be long term solution. Is sufficient for simple libc calls until we have better override support for crucible.
   - Added PLT stubs for `malloc` and `clock` for Aarch32
   - No PPC support yet (see: https://github.com/GaloisInc/pate/issues/327). Test marked as expected failure.
-  Clean up build system for tests.
   - Factor out duplication between architectures


src/Pate/Config.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 module Pate.Config (
   PatchData(..),
+  PatchDataParseError,
```
travitch, line 6: Should this have its constructors exported too?

src/Data/Macaw/CFGSlice.hs
```diff
@@ -126,15 +128,25 @@ assignBlockEnd archFns blendVar stmt = MSB.crucGenArchConstraints archFns $ do
   blend' <- blockEndAtom archFns blend
   MSB.addStmt $ CR.WriteGlobal blendVar blend'
 
+blockEndCaseEq :: IsSymInterface sym
+               => Proxy arch
```
travitch, line 132: It is generally a best practice to make proxies in positive (argument) positions type variables:

```
blockEndCaseEq :: forall sym arch proxy => IsSymInterface sym -> proxy arch -> ...
```

This way the caller can pass in any valid value of the correct kind (given that this function doesn't care what is passed in).

src/Pate/Config.hs
```diff
@@ -202,6 +204,16 @@
...
 
+-- | Controls how abstract domains are shared between function call nodes in the pairgraph.
+data ContextSensitivity =
+    -- | All calls to any given function share the same abstract domain, i.e.
+    --   the resulting domain is proven to hold at every return site.
+    AllSharedAbstractDomains
+    -- | Each call to a function has a distinct abstract domain that holds
+    --   at its (single) return site (distinguished by tagging the function node with
+    --   the calling node)
+  | AllDistinctAbstractDomains
```
travitch, line 215: "All" is not quite correct, is it? What is the key in the map from contexts to abstract values? What happens in the presence of recursion?

src/Data/Parameterized/SetF.hs
```diff
@@ -26,20 +26,34 @@
...
-import Prelude hiding (filter)
+import Prelude hiding (filter, null)
 import           Data.Parameterized.Classes
+import qualified Data.Foldable as Foldable
+
+import qualified Prettyprinter as PP
+import           Prettyprinter ( (<+>) )
+
+import qualified Prettyprinter as PP
+import           Prettyprinter ( (<+>) )
```
travitch, line 56: These imports are duplicated

src/Pate/AssumptionSet.hs
```diff
@@ -0,0 +1,340 @@
...
+
+import           Data.Parameterized.SetF ( SetF )
+import qualified Data.Parameterized.SetF as SetF
+import qualified Pate.SimulatorRegisters as PSR
+import           Pate.Panic
+import qualified What4.ExprHelpers as WEH
+import           What4.ExprHelpers ( ExprSet, VarBindCache, ExprBindings, ppExprSet )
+import qualified Pate.ExprMappable as PEM
+
+data AssumptionSet sym =
```
travitch, line 57: Is there some kind of invariant that this type is attempting to capture on top of `ExprSet`?

src/Pate/ExprMappable.hs
```diff
@@ -132,3 +140,21 @@ instance ExprMappable sym (SkipTransformation a) where
...
+-- | Deduce ad-hoc 'ExprMappable' instances for 'WI.SymExpr'.
+--   This uses an unsafe coercion to implicitly use the 'ExprMappable' instance
+--   defined in 'ToExprMappable'.
+--   This is necessary because 'WI.SymExpr' is a type family and therefore we cannot
+--   declare the obvious instance for it.
+symExprMappable ::
+  forall sym.
+  sym ->
+  SymExprMappable sym (WI.SymExpr sym)
+symExprMappable _sym = unsafeCoerce r
```
travitch, line 158: This is quite scary.

Why is this safe? Is there no other way to achieve what you need here?

src/What4/ExprHelpers.hs
```diff
@@ -115,10 +117,69 @@
...
+            let rng_y = W4.integerBounds y
+            W4.unsafeSetAbstractValue (W4AD.rangeMax rng_x rng_y) <$>
+              W4.intIte sym x_le_y y x
+
+-- | Overriding 'W4.integerToNat' to use a modified 'intMax' that
+--   statically evaluates in more cases. In particular this ensures that
+--   integers with a positive abstract domain aren't wrapped in any additional checks,
+--   and that the inner integer is statically known to be positive.
+integerToNat ::
+  W4.IsExprBuilder sym =>
```
travitch, line 170: I'm obviously always worried about unsafeCoerce. Is there a time horizon to replace this with safe what4 alternatives now that we have improved infrastructure there? If so, can you reference the ticket here?

src/Pate/Arch.hs
```diff
@@ -127,12 +143,80 @@ data ValidArchData arch =
                 --
                 -- For example, these could be PLT stub symbols for ELF binaries
                 , validArchPatchedExtraSymbols :: Map.Map BS.ByteString (BVS.BV (MC.ArchAddrWidth arch))
+                , validArchStubOverrides :: ArchStubOverrides arch
                 }
 
+-- | A PLT stub is allowed to make arbitrary modifications to the symbolic state
```
travitch, line 149: This is currently used for PLT stubs, but could in principle be used for any function that is to be overridden. That doesn't really require any changes, but might be worth a note in the comment.

src/Pate/Arch.hs
```diff
@@ -127,12 +143,80 @@
...
+mkMallocOverride _rLen rOut = StubOverride $ \sym st -> do
+  let mr = PS.simMaxRegion st
+  let w = MC.memWidthNatRepr @(MC.ArchAddrWidth arch)
+  mr_nat <- W4.integerToNat sym (PS.unSE mr)
+  zero <- W4.bvLit sym w (BVS.mkBV w 2)
+  let fresh_ptr = PSR.ptrToEntry (CLM.LLVMPointer mr_nat zero)
+  mr_inc <- PS.forScopedExpr sym mr $ \sym' mr' -> do
+    one <- W4.intLit sym' 1
+    W4.intAdd sym' mr' one
+  return (st { PS.simMaxRegion = mr_inc, PS.simRegs = ((PS.simRegs st) & (MC.boundValue rOut) .~ fresh_ptr) })
```
travitch, line 183: Can you add a comment that describes the semantics of the override, perhaps with a justification for the behavior?

src/Pate/Discovery.hs
```diff
@@ -269,8 +256,9 @@ associateFrames ::
   forall sym arch v.
   SimBundle sym arch v ->
   MCS.MacawBlockEndCase ->
-  EquivM sym arch (PSS.AssumptionSet sym v)
-associateFrames bundle exitCase = PPa.catBins $ \get -> do
+  Bool ->
```
travitch, line 259: At the very least we should document what this boolean is. Better, we can make it into an ADT that is descriptive at the call site

src/Pate/Arch.hs
```diff
@@ -127,12 +143,80 @@
...
+  16 <= MC.ArchAddrWidth arch =>
+  MS.SymArchConstraints arch =>
+  MC.ArchReg arch (MT.BVType (MC.ArchAddrWidth arch)) {- ^ return register -} ->
+  StubOverride arch
+mkClockOverride rOut = StubOverride $ \sym st -> do
+  let w = MC.memWidthNatRepr @(MC.ArchAddrWidth arch)
+  zero_bv <- W4.bvLit sym w (BVS.mkBV w 0)
+  zero_nat <- W4.natLit sym 0
+  let zero_ptr = PSR.ptrToEntry (CLM.LLVMPointer zero_nat zero_bv)
+  return (st { PS.simRegs = ((PS.simRegs st) & (MC.boundValue rOut) .~ zero_ptr) })
```
travitch, line 196: Likewise, can you document the semantics of the override?

src/Pate/Equivalence.hs
```diff
@@ -444,27 +449,29 @@ regCondToAsm ::
...
   RegisterCondition sym arch v ->
-  IO (AssumptionSet sym v)
+  IO (AssumptionSet sym)
 regCondToAsm _sym regCond = return $ PRt.collapse (regCondPreds regCond)
 
 -- | A structured pre condition
 data StatePreCondition sym arch v = StatePreCondition
   { stRegPreCond :: RegisterCondition sym arch v
   , stStackPreDom :: PEM.MemoryDomain sym arch
   , stMemPreDom :: PEM.MemoryDomain sym arch
```
travitch, line 459: What should I (the reader) expect to appear in an "extra" assumption set?

src/Pate/Equivalence.hs
```diff
@@ -476,13 +483,20 @@
...
-eqDomPre sym inO inP (EquivContext hdr _stackRegion) eqDom  = do
+eqDomPre sym inO inP (eqCtxHDR -> hdr) eqDom  = do
   let
     stO = simInState inO
     stP = simInState inP
     
   regsEq <- regDomRel hdr sym stO stP (PED.eqDomainRegisters eqDom)
-  return $ StatePreCondition regsEq (PED.eqDomainStackMemory eqDom) (PED.eqDomainGlobalMemory eqDom)
+  -- TODO: we haven't included this in the domain, and therefore we're simply
+  -- requiring that the number of allocations for each program always match
```
travitch, line 493: Can you open an issue to track this?

src/Pate/AssumptionSet.hs
```diff
@@ -0,0 +1,340 @@
...
+macawRegBinding _sym var val = do
+  case PSR.macawRegRepr var of
+    CLM.LLVMPointerRepr _ -> ptrBinding (PSR.macawRegValue var) (PSR.macawRegValue val)
+    CT.BoolRepr -> exprBinding (PSR.macawRegValue var) (PSR.macawRegValue val)
+    CT.StructRepr Ctx.Empty -> mempty
+    _ -> panic Rewriter "macawRegBinding" ["Unexpected macaw types"
+                                          , show (PSR.macawRegRepr var)
+                                          , show (PSR.macawRegRepr val)
+                                          ]
+
```
travitch, line 197: These would all benefit from saying something about what the meaning of the result is (and hopefully that is in terms of the description of `AssumptionSet`)

src/Pate/AssumptionSet.hs
```diff
@@ -0,0 +1,340 @@
...
+  m f
+apply sym asm f = do
+  cache <- IO.liftIO WEH.freshVarBindCache
+  apply' sym cache asm f
+
+-- | Rewrite the given 'f' with any bindings in the given 'AssumptionSet'.
+--   Bindings are applied repeatedly to each component expression until a fixpoint
+--   is reached or a loop is detected.
+apply' ::
+  forall sym f m t solver fs.
```
travitch, line 220: Do both of these need to be exported?

src/Pate/AssumptionSet.hs
```diff
@@ -0,0 +1,340 @@
...
+  PEM.ExprMappable sym f =>
+  IO.MonadIO m =>
+  sym ~ (W4B.ExprBuilder t solver fs) =>
+  sym ->
+  VarBindCache sym ->
+  AssumptionSet sym ->
+  f ->
+  m f
+apply' sym cache asm f = do
+  let
```
travitch, line 230: If this does need to be exported, a more descriptive name like `applyWithCache` would be better than `apply'`---understanding what the prime means in isolation is difficult

src/Pate/AssumptionSet.hs
```diff
@@ -0,0 +1,340 @@
...
+  cache <- IO.liftIO WEH.freshVarBindCache
+  origAsm' <- PEM.mapExpr sym (apply' sym cache newAsm) origAsm
+  return $ newAsm <> origAsm'
+
+-- | Retrieve a value that the given expression is bound to in
+--   the given 'AssumptionSet'.
+--   In the case of multiple results, constant values are preferred,
+--   otherwise the least element of all possible symbolic bindings is returned.
+getSomeBinding ::
+  forall sym tp.
```
travitch, line 260: Can you say something about why one might want this?

src/Pate/AssumptionSet.hs
```diff
@@ -0,0 +1,340 @@
...
+      in SetF.unions (map (\e' -> SetF.insert e' (getAllBindings sym ancestors' asm e')) (SetF.toList es))
+    Nothing -> case W4.exprType e of
+      W4.BaseBoolRepr | isAssumedPred asm e -> SetF.singleton (W4.truePred sym)
+      _ -> SetF.empty
+
+-- | Compute a predicate that collects the individual assumptions in the frame, including
+-- equality on all bindings.
+toPred ::
+  forall sym m.
+  W4.IsSymExprBuilder sym =>
```
travitch, line 296: Can you state precisely when the returned predicate is true? Is it true if all of the assumptions hold if you assume the initial equalities on bindings?

travitch:
I did my best here given the size. If there is anything you'd like me to focus on more closely, let me know and I will make an attempt

src/Pate/AssumptionSet.hs
danmatichuk: At the moment, no, but I had some ambitions to use it later.

src/Pate/AssumptionSet.hs
danmatichuk: I don't think there are any explicit invariants here. The assumption set is a collection of predicates for general predicates as well as bindings used for equality assumptions that can be applied to perform simplifications.

The state of this binding environment may need to be revisited at some point to ensure that expressions can be consistently simplified into some kind of normal form (which would imply some kind of invariant on what shape the bindings in the environment are allowed to have). At the moment the rewriter that consumes the bindings doesn't make any assumptions about their shape (i.e. it implicitly breaks rewrite loops while trying to find a rewriting fixpoint).

PR #361, opened 2024-02-15
Push docker containers to Artifactory

dwest-galois:
Reference Ticket #2868

I generated an Access Token that will allow the actions build to push docker containers to the `pate_docker-local`
repository on Artifactory. This token has been stored in the Github secrets, that can be accessed from the CI job.

$ARTIFACTORY_USER: This is just the tokens name (pate_ci)
$ARTIFACTORY_KEY: This is the actual access token that was generated

For the tag, Github only has built-in support for the long version of the commit hash. I went ahead and added a step to convert it to the short version, which should make it easier to read and reference. So now, you can pull the pate images from the following:

artifactory.galois.com:5025/pate/pate:$CI_COMMIT_SHORT_SHA

I standardized on the Gitlab naming convention of "$CI_COMMIT_SHORT_SHA" for the SHORT_SHA. If you think that SHORT_SHA will be problematic for any reason, let me know and I can switch it back to the $GITHUB_SHA.

It looks like it successfully finished and the container appears to be here:

[https://artifactory.galois.com:443/artifactory/pate_docker-local/pate/pate/496fa37/](https://artifactory.galois.com/artifactory/pate_docker-local/pate/pate/496fa37/)

Can I have you confirm that everything looks correct to you?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ name: CI Matrix
 on:
   pull_request:
   push:
-    branches: [master]
+    branches: [master, dwest/test-docker]
 
 jobs:
   docker:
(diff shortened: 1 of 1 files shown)


danmatichuk:
Looks good to me, however I'm wondering if it'd make sense for the script to pull the layers from artifactory before it builds in order to save time. The Dockerfile is written to make use of caching for intermediate layers in order to skip some of the more expensive build steps if it can.

danmatichuk (approved):

PR #381, opened 2024-03-26
Dm/trace trees

danmatichuk:
add SymSequenceTree datatype for representing all possible traces through a CFAR


jim-carciofini:
@danmatichuk I think this is good to pull into master assuming CI passes. I have the GUI processing the instruction trees and new dialogs to display the instruction trees.

PR #412, opened 2024-06-21
binja: explicitly specify expected platform for pate container

thebendavis:
Match platform in Dockerfile, avoid platform mismatch warning on non-native platforms (e.g. arm64 macs)


jim-carciofini (approved):
Looks good. Same docker arg I used for MCAD.

PR #417, opened 2024-07-09
Add packet example and integration test

thebendavis:
This adds a standalone example (`packet.exe`) and an integration test. Beyond testing basic verifier functionality, this is intended to provide an integration test of the 1) terminal user interaction, and 2) resulting equivalence condition, in ways that the other automated tests do not. The integration test is implemented as an `expect` script that captures a subset of the expected output including a key clause of the equivalence condition, and will alert / need to be updated if this output changes during evolution of the PATE verifier.

This integration test has been added to the `docker` CI job. I have not additionally added it to the `build` CI job because that job uses `cabal build pkg:pate` which breaks subsequent use of `./pate.sh`, which is the interface used by the integration test. This issue is captured separately in #418 and is outside of the scope of this PR.


danmatichuk:
This looks fine to me, I think including at least one example that bypasses the normal "scripting" interface makes sense as a sanity check. I think it also makes sense to test it both via docker and natively once #418 is addressed.

.github/workflows/main.yml
```diff
@@ -54,6 +54,10 @@ jobs:
           cat out.log
           grep -e "All .* tests passed" out.log || exit 1
 
+      - name: Docker Integration Test
```
danmatichuk, line 57: We could consider just using this to replace the above integration test, which only exists as a minimal test to ensure that the verifier actually works in the container.

tests/integration/packet/packet.exp
```diff
@@ -0,0 +1,72 @@
...
+
+set timeout 480
+
+expect_before {
+        timeout { send_user "\n\nFAILURE: timeout\n"; exit 1 }
+        eof     { send_user "\n\nFAILURE: eof\n";     exit 2 }
+}
+
+expect "Choose Entry Point"
+expect "1: Function Entry \"parse_packet\" (segment1+0x554)"
```
danmatichuk, line 28: One of the main motivations for the scripting interface was to avoid referencing specific numbers from the menus. I'm assuming that's not something we can easily do here?

tests/integration/packet/packet.exp
```diff
@@ -0,0 +1,72 @@
...
+
+expect "0: segment1+0x644 \\\[ via: \"parse_packet\" (segment1+0x554) \\\]"
+expect ">"
+send "0\r"
+
+expect "3: let -- segment1+0x664.. in not v40591"
+expect ">"
+send "3\r"
+
+expect "v40591 = and (eq 0x80:\\\[8\\\] v40584) (not (eq v40587 (bvSum v40584 0x80:\\\[8\\\])))"
```
danmatichuk, line 61: We probably want to figure out how to avoid making this depend on the exact variable identifiers (e.g. v40591) since those will almost certainly change fairly regularly.

jim-carciofini: If we pulled the simplification logic from the GUI into Pate, would that fix this issue? I believe want to do that anyway as time permits.

thebendavis: Agreed - this will be relatively brittle for now so we'll have to replace this line when changes to the verifier result in changes to this expression. Ultimately we'd like to match not on this low-level view of the equivalence condition expression but instead on the simplified version we show to operators, so we could match something like:

```
(#x80 != read1#00011045)) | (read1#00011044) = (#x80 + read1#x00011045)))
```

but currently the simplification logic for producing this expression spans both the verifier and the Binary Ninja plugin. I've opened issue #420 about unifying this logic in the verifier, once that is resolved it will natural to update our regression test to leverage it.

tests/integration/packet/packet.exp
thebendavis: It's definitely an option to, for example, regex parse the lines of menus from PATE to get the appropriate number to send in response to each step. That would reduce some of the brittleness due to menu item ordering, but we'd still have either some level of brittleness or some level of ambiguity matching the rest of each line depending on how strictly we check for things.

My thinking here is that we have the rest of the PATE scripts which do the more structural matching for the other test cases - for this integration test, given that it'll still be brittle in ways beyond just the menu numbers, we might prioritize keeping the expect script as simple as possible and just use the basic string match. Doing it this way means that if/when we need to make updates to the expect script, it's just a matter of copy-pasting in the expected line, rather than adapting or writing a new regex expression to capture the appropriate context. I think that's an ok tradeoff for this single small example - if this ends up flagging too many issues that *aren't* interesting to us, or if we later end up doing lots more / longer integration tests like this, then I think we may want to explore something that does more sophisticated parsing of PATE output.

.github/workflows/main.yml
thebendavis: Sure, I'm ok with that - this will save a few minutes off of the `docker` build job in CI, and we're still running tests in the `build` job so we shouldn't miss much. Done.

thebendavis:
> This looks fine to me, I think including at least one example that bypasses the normal "scripting" interface makes sense as a sanity check. I think it also makes sense to test it both via docker and natively once #418 is addressed.

Thanks for the review! I have flagged the appropriate items for followup in relevant issues.

PR #442, opened 2024-09-03
change 'no-assume-stack-scope' to 'assume-stack-scope'

danmatichuk:
This flips the default behavior of this unsafe flag.

If set, 'assume-stack-scope' injects an additional assumption into every pre-domain that all out of
scope stack values (i.e. past the stack pointer)

This is a fairly strong assumption that is unsafe in general, in particular it will crash in the case where it is provably false (i.e. the value domain implies that the stack contents past the stack pointer are necessarily inequivalent)


docs/user-manual/invoking.tex
```diff
@@ -88,8 +88,8 @@ \section{Invoking the \pate{} Verifier}
   --read-only-segments ARG Mark segments as read-only (0-indexed) when loading
                            ELF
   --script FILENAME        Save macaw CFGs to the provided directory
-  --no-assume-stack-scope  Don't add additional assumptions about stack frame
-                           scoping
+  --assume-stack-scope     Add additional assumptions about stack frame 
+                           scoping during function calls (unsafe)
```
thebendavis, line 92: While we're here, is there a brief description of what exactly this does and when the user would want to use it that we can add to the user manual? Maybe just a bit of elaboration at the end of `invoking.tex` framed as tips for invoking PATE, perhaps?

